PR #8, opened 2019-05-15
Add vendor dependencies

lluiscampos:
@kacf Please have a look


lluiscampos:
Good, because I don't intent to auto-deploy it ;) But thanks for the recipe

This PR was only to get a base where to start my work, and because it couldn't compile anymore with latest upstream libraries.

PR #9, opened 2019-05-20
QA-74 add gitlab pipelines trigger

lluiscampos:
QA-74: Replace Jenkins' build trigger for GitLab ones

Two main changes:
- GitLab will not run Pipelines for PRs coming from forks. To workaround
this and continue with our current workflow, we will be creating
branches like pr_XXX for each incoming PR on a given microservice or
component.
- Modify the main triggerBuild logic to access GitLab API instead of
Jenkins. We use the same names for the build parameters. As per now the
Pipeline itself will do nothing, but the bot is ready for QA-75


lluiscampos:
@kacf Please have a look!

You can see it "working" on [this inventory PR](https://github.com/mendersoftware/inventory/pull/131) and the main [Mender QA Pipeline](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines) triggers.

main.go
```diff
@@ -138,14 +134,21 @@ func main() {
...
+			// coming from a mendersoftware member or not (equivalent to the old Travis tests)
+			err := createPullRequestBranch(*pr.Repo.Name, strconv.Itoa(pr.GetNumber()), action)
+			if err != nil {
+				log.Errorf("Could not create PR branch: %s", err.Error())
+			}
 
+			// Then, continue to the integration Pipeline only for mendersoftware members
 			if member, _, _ := githubClient.Organizations.IsMember(context, "mendersoftware", pr.Sender.GetLogin()); !member {
-				log.Warnf("%s is making a pullrequest, but he's not a member of mendersoftware, ignoring", pr.Sender.GetLogin())
+				log.Warnf("%s is making a pullrequest, but he/she is not a member of mendersoftware, ignoring", pr.Sender.GetLogin())
```
kacf: :+1:

main.go
```diff
@@ -286,26 +285,26 @@ func triggerBuild(conf *config, build *buildOptions) error {
 				return err
 			} else {
 				log.Infof("%s version %s is being used in %s: ", versionedRepo, version, build.baseBranch)
-				buildParameter.Add(repoToJenkinsParameter(versionedRepo), version)
+				buildParameters = append(buildParameters, &gitlab.PipelineVariable{repoToJenkinsParameter(versionedRepo), version})
```
kacf: Maybe rename `repoToJenkinsParameter` to `repoToParameter`?

main.go
```diff
@@ -161,7 +164,7 @@ func main() {
 			}
 		}
 	})
-	r.Run("0.0.0.0:8081")
+	r.Run("0.0.0.0:8083")
 }
 
 func parsePullRequest(conf *config, action string, pr *github.PullRequestEvent) []buildOptions {
```
lluiscampos: It seems to me that the logic in this function to decide on publishArtifacts build parameter is broken, as `buildContainers` will never be true. What is the historical reason for this? Should I fix or clean it up?

main.go
lluiscampos: I renamed it to `repoToBuildParameter`

main.go
kacf: Just clean it up. IIRC it was an attempt at auto-pushing containers, but it never worked as intended.

lluiscampos: Thanks, I cleaned it up. I leave it as an independent commit so that it can be reviewed if we ever need this feature properly implemented.

lluiscampos:
Thanks for your review! I am merging and preparing the systemd service in Jenkins master to run this version in addition to the current one.

PR #10, opened 2019-06-21
Add functionality for disabling integration tests in a PR body

oleorhagen:
Changelog: Title

### DIFF
--- a/main.go
+++ b/main.go
@@ -168,6 +168,11 @@ func parsePullRequest(conf *config, action string, pr *github.PullRequestEvent) 
 	log.Info("Pull request event with action: ", action)
 	var builds []buildOptions
 
+	// Do not run the integration tests if '*NO-INTEGRATION*' is found on a separate line in the commit body
+	if pr.Body != nil && strings.Contains(pr.Body, "NO-RUN-TESTS") {
+		return nil
+	}
+
 	repo := *pr.Repo.Name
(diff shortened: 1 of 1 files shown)


oleorhagen:
@kacf Unfamiliar territory for me, but I think this should do the trick

oleorhagen:
The keyword is up for grabs btw

main.go
```diff
@@ -168,6 +168,11 @@ func parsePullRequest(conf *config, action string, pr *github.PullRequestEvent)
 	log.Info("Pull request event with action: ", action)
 	var builds []buildOptions
 
+	// Do not run the integration tests if '*NO-INTEGRATION*' is found on a separate line in the commit body
+	if pr.Body != nil && strings.Contains(pr.Body, "*NO-INTEGRATION*") {
```
kacf: `strings.Contains` doesn't accept wildcards. Just the string alone should be enough.

I would also consider naming it `NO-RUN-TESTS`, since this is not limited to integration tests; acceptance tests also will benefit from this.

kacf:
It will also need to be compiled/restarted on Jenkins. Let me know if you don't know the procedure.

main.go
oleorhagen: They weren't wildcards :p It was supposed to be a part of the marker.

oleorhagen:
@kacf changed the test-marker to suit your taste.

PS: I want to know the procedure.

PR #13, opened 2019-07-23
Revert Gitlab functionality

kacf:
We submitted the Gitlab refactor too soon, we still need the Jenkins one.


kacf:
@mzedel, @lluiscampos: FYI.

I have put the necessary commits in the "gitlab" branch, so we can easily reapply them later.

PR #17, opened 2019-09-04
Getting GitLab branch into master

lluiscampos:
As GitLab is our now default CI server, we are bringing this branch to master and leave a newly created "jenkins" branch for future back-ports.


lluiscampos:
@kacf please a quick check on this.

Also, should we then delete gitlab branch to avoid confusion?

kacf:
Yep, merge and get rid of the branch.

kacf:
You can even consider getting rid of the `jenkins` branch, unless we have any modifications there. It is present in history, so it does not really provide any value, and we have already turned it off.

PR #19, opened 2019-09-24
Stop treating tenantadm specially and treat as release repo.

kacf:
Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>


kacf:
@lluiscampos: This is already deployed. I'm not 100% sure that this will work correctly with older branches, but let's cross that bridge when we get there. Master is most important right now.

PR #23, opened 2019-11-08
Fix: Build PR's moved out of the 'Draft' status

oleorhagen:
The Github PR's will not get the status 'opened', like a regular PR when moved
out of the 'Draft' state. However, it gets the status 'ready_for_review', which
currently is not a recognized action in the integration-test-runner.

This fix adds the 'ready_for_review' status to the accepted status for builds.

Changelog: None

### DIFF
--- a/main.go
+++ b/main.go
@@ -205,7 +205,8 @@ func parsePullRequest(conf *config, action string, pr *github.PullRequestEvent) 
 	commitSHA := pr.PullRequest.Head.GetSHA()
 
 	// github pull request events to trigger a jenkins job for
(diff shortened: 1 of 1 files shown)


oleorhagen:
Missed in the original merge for QA-107. Appareantly the status is different when a PR is moved out of the Draft stage, than when it is created the first time.

oleorhagen:
Verified in: https://github.com/mendersoftware/mender/pull/451

oleorhagen:
Probs should have been caught the first time around, but it did not occur to me that it would have it's own independent status for moving out of review. Anywho, good to go now.

PR #24, opened 2019-11-14
[Experimental] Move from govendor to gomod

oleorhagen:
Initial experimentation with go mod.

So far it seems pretty nice, although quite different from the govendor
workflow.


oleorhagen:
How about this?

lluiscampos:
I don't know about go mod. What are the advantages over govendor? Are you (we) thinking about changing it in the core repos?

oleorhagen:
It is now officially a part of the language, and as it is now supported in all golang versions we use (1.11-13), moving is fine, and I think that this should be considered at some point, even though it is a little bit of work (but not too bad I think). 

The go module setup is a little different than we currently do it. In general a module is a freestanding project. This means that we could get rid of vendoring all together! a go build command with a `go.mod` file will pull in the correct dependencies. 

This also means that GOPATH is not strictly enforced any longer. The project is now freestanding! 

In fact, now, with the vendor folder missing (no  dependencies come with the project by default, although vendoring can be enabled). The go build command will pull in the correct versions of the dependencies from the go.mod file, as you build it! (pretty neat imo!).

It also enables multiple versions of a library in the same project (through semantic import paths, which will be the go standard going forward).

I think that this is something we are going to have to do sooner or later, no matter what.

lluiscampos:
Thanks for the explanation. I can see that we will have this "sooner or later" everywhere. But I still think we need to discuss it in the team.

One of the details that I love with govendor is how easy it is to do git bisect, for example. I understand that go mod the workflow will still be possible, but I am skeptical on how will they maintain old references to projects, or directly projects that have been removed.

In the other hand, what should I really review here? Maybe @mzedel can do the actual review...

oleorhagen:
> Thanks for the explanation. I can see that we will have this "sooner or later" everywhere. But I still think we need to discuss it in the team.
> 
Yes, most definitely. I have no plans on adding this to our 'main' repositories at the moment. But I figured this was a nice fringe repo to let people have a look at it, and perhaps give go mod a trial run.

> One of the details that I love with govendor is how easy it is to do git bisect, for example. I understand that go mod the workflow will still be possible, but I am skeptical on how will they maintain old references to projects, or directly projects that have been removed.
> 
I'm not quite sure I understand this one. The builds are still reproducible?
Are you asking how the dependencies are tracked? If so, they are committed with the `go.mod` file, and the `go.sum` file, which is sort of the equivalent of the `govendor.json` file we are using now. So it should still be traceable.

> In the other hand, what should I really review here? Maybe @mzedel can do the actual review...

Whomever. Still, we need to decide if we want to give this a trial run or not.

lluiscampos:
Thanks. I follow your reasoning and:

> Whomever. Still, we need to decide if we want to give this a trial run or not.

Yes! Let's give it a try!

PR #25, opened 2019-12-05
MEN-2892 mender-helm repo added

merlin-northern:
ChangeLog:title
Signed-off-by: Peter Grzybowski <peter@northern.tech>

### DIFF
--- a/main.go
+++ b/main.go
@@ -86,6 +86,7 @@ func getConfig() (*config, error) {
 			"tenantadm",
 			"deployments-enterprise",
 			"useradm-enterprise",
+			"mender-helm",
 		}
 
 	watchRepositories := os.Getenv("WATCH_REPOS")


merlin-northern:
> @merlin-northern can you please use tabs instead of spaces?

as a rule I always use spaces. but lets be consistent. fixed. thanks.

merlin-northern:
> This change is unnecessary. Sorry I miss lead to this.
> 
> If you look at the code, you can see that this list of repos is only relevant for triggering the integration pipeline. The repo at hand, `mender-helm` is independent and only needs to trigger the single component pipeline, which is handled in function `createPullRequestBranch` regardless of `defaultWatchRepositories` list.

ok no worries! I remember adding other repos here, so I thought it is fine. right. closing this one.

lluiscampos:
> ok no worries! I remember adding other repos here, so I thought it is fine. right. closing this one.

You remember right. You added `useradm-enterprise` which is part of the integration Pipeline. So as a rule of thumb, only repos for backend core microservices need to be added to the list

PR #29, opened 2020-03-17
QA-171: Redo function to use --in-integration-version

lluiscampos:
The previous approach will not work when adding new repos: these will be
listed in master but not present in integration/2.x.x branches.

The original reason to do this clones is unknown.

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -2,8 +2,6 @@ package main
 
 import (
 	"fmt"
-	"io/ioutil"
-	"os"
 	"os/exec"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


lluiscampos:
Thanks @oleorhagen  for finding the bug! I only committed here what you discovered :)

helpers.go
```diff
@@ -2,39 +2,23 @@ package main
...
-
-	gitcmd := exec.Command("git", "clone", "-b", baseBranch, "https://github.com/mendersoftware/integration.git", tmp)
-
-	// timeout and kill process after GIT_OPERATION_TIMEOUT seconds
-	t := time.AfterFunc(GIT_OPERATION_TIMEOUT*time.Second, func() {
-		gitcmd.Process.Kill()
-	})
-	defer t.Stop()
+func getServiceRevisionFromIntegration(repo, baseBranch string, conf *config) (string, error) {
+	c := exec.Command("release_tool.py", "--version-of", repo, "--in-integration-version", baseBranch)
```
kacf: Note that you will still need to keep pulling the repository to keep it up to date. I'm not sure if that is handled now that this portion is gone.

lluiscampos: Yep. It is done once per PR here: https://github.com/mendersoftware/integration-test-runner/pull/29/files#diff-01127be9ed6b6352616a2f1c10bc3540R24

helpers.go
```diff
@@ -2,39 +2,23 @@ package main
...
+	c.Dir = conf.integrationDirectory + "/extra/"
+	version, err := c.Output()
 
-	if err := gitcmd.Run(); err != nil {
-		return "", fmt.Errorf("failed to 'git clone' the integration folder: %s\n", err.Error())
+	if err != nil {
```
oleorhagen: Why not return `strings.TrimSpace(string(version)), err` ?

oleorhagen:
> The previous approach will not work when adding new repos: these will be
> listed in master but not present in integration/2.x.x branches.
> 
> The original reason to do this clones is unknown.

The best kinds of reasons

PR #31, opened 2020-03-30
Support for different GitLab sumgroups

lluiscampos:
Up until now, all our repos have been ported to "Mender" subgroup, but
from now on we will have some repos in different groups (aka SaaS).
    
Let the code define which repos go to a different subgroup.

### DIFF
--- a/main.go
+++ b/main.go
@@ -43,6 +43,10 @@ type buildOptions struct {
 	makeQEMU   bool
 }
 
+var githubRepoToGitLabProject = map[string]string{
+	"saas": "Northern.tech/MenderSaaS/saas",
+}
+
 const (
 	GIT_OPERATION_TIMEOUT = 30
 )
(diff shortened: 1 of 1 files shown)


lluiscampos:
I deployed the fix to production and "it works".

main.go
```diff
@@ -43,6 +43,10 @@ type buildOptions struct {
 	makeQEMU   bool
 }
 
+var githubRepoToGitLabProject = map[string]string{
```
oleorhagen: Thoughts:
Maybe we should hard-code them all, while we're at it?
I understand the lure of assuming `Mender` on all of them, but it does mean that all new ones will slip into this group.
Assuming `Mender` is the least privileged, this isn't really an issue though.

I assume you've been through this already, what did you end up with?

lluiscampos: I see your point, but hard-coding them will force us to update the code here every single time that we create a repo. As per today, we "only" have to update the code if the new repo is part of the Mender base repositories (i.e. needs to trigger the integration pipeline).

PR #33, opened 2020-04-14
QA-189: Syncronize GitHub/GitLab branches and tags for private repos sync branches

lluiscampos:


lluiscampos:
I have it running on a separate hook in the server (using a different port) and it seems to synchronize the branches correctly. I have not verified the tags part, so please pay attention there to the git commands.

The only issue is that it will not work when deleting refs. I thought about it but decided that it is not worth the effort, as we should not care much if GitLab git repo has old unused branches laying around.

main.go
```diff
@@ -49,6 +49,31 @@ var githubRepoToGitLabProject = map[string]string{
...
+	"saas",
+	"saas-tools",
+	"sre-tools",
+	// docs
+	"mender-docs",
+	"mender-docs-site",
+	// qa
+	"mender-qa",              // OS in mendersoftware
+	"mender-test-containers", // OS in mendersoftware
+}
```
kacf, line 75: I don't understand the selection of repositories in this list. Why are some, but not all of the OS repositories listed?

lluiscampos: We need to sync the project that are private in GitLab, regardless of their status in GitHub. Only some OS repositories have private projects in GitLab.

I think that except for `mender-qa`, we could open the other projects in GitLab.

kacf: Yes, fully agree, open as many as possible. Just be careful that the repo is not exposing any credentials.

Please add a comment about the Github/Gitlab difference, since this was not obvious from the list.

lluiscampos: Fixed now.

main.go
```diff
@@ -75,8 +75,10 @@ func getConfig() (*config, error) {
 	}
 
 	// if no env. variable is set, this is the default repo watch list
+	// This list must be kept in sync with: release_tool.py --list git
```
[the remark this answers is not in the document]
lluiscampos: Hehehe. This I gave it a though, actually. What stopped me is that the configuration is taken once, while this (getting the repo list from the tool) should be done in the loop after every pull of integration repo (we pull all the time to have latest master).

I have it in my paper org as a possible value add follow-up... But I think we should merge this as is, so that we get the fix in place ASAP.

PR #40, opened 2020-07-03
Fixes for image in kubernetes

lluiscampos:
* Listen for the hook in root instead of incoming
As the tool will have a dedicated domain when deployed in the new infra.

* Add missing dependencies and initialization to the image
The image was missing:
    * openssh
    * SSH client initialization
    * release tool dependencies (pyyaml)
    * release tool in path

* Commented out in docker-compose file how the SSH keys are expected to be
mounted in the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ ADD ./ .
 RUN CGO_ENABLED=0 go build
 
 FROM alpine:3.11
-EXPOSE 8083
+EXPOSE 8080
 RUN apk add git openssh python3 py3-pip
 RUN pip3 install --upgrade pyyaml
 RUN mkdir -p /root/.ssh/ && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,10 +6,16 @@ RUN CGO_ENABLED=0 go build
 
 FROM alpine:3.11
 EXPOSE 8083
-RUN apk add git
+RUN apk add git openssh python3 py3-pip
+RUN pip3 install --upgrade pyyaml
+RUN mkdir -p /root/.ssh/ && \
+    git config --global user.name "Mender Root" && \
+    git config --global user.email root@mender-jenkins.mender.io
 RUN git clone https://github.com/mendersoftware/integration.git /integration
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -175,7 +175,7 @@ func main() {
 	githubClient := createGitHubClient(conf)
 	r := gin.Default()
 
-	r.POST("/incoming", func(context *gin.Context) {
+	r.POST("/", func(context *gin.Context) {
```
0lmi, line 178: Currently as workaround I configured nginx reverse proxy to replay 200 on GET to `/` endpoint. This is needed for google load balancer to determinate if service is healthy. If LB doesn't get 200 replies then it doesn't forward traffic to the container. Would be great to have this as a part of test runner bot itself.

Dockerfile
```diff
@@ -6,9 +6,15 @@ RUN CGO_ENABLED=0 go build
 
 FROM alpine:3.11
 EXPOSE 8083
-RUN apk add git
+RUN apk add git openssh python3 py3-pip
+RUN pip3 install --upgrade pyyaml
+RUN mkdir -p /root/.ssh/ && \
+    ssh-keyscan github.com >> /root/.ssh/known_hosts && \
```
0lmi, line 12: Good stuff, but, I'm just thinking, might it be that github changes IPs from time to time. If it's so then we'll need to rebuild an image when this happens. But if we add this into entrypoint then only container restart will require to handle such case.

main.go
lluiscampos: I don't understand: do you want me to revert this change?

I know that you configured nginx to handle this, but I think we can now remove that handling and just let the container reply

Dockerfile
lluiscampos: Good point. I'll move it to a new entrypoint script.

lluiscampos: Done.

main.go
lluiscampos: Now I understood, thanks. Fixing...

lluiscampos: Done :+1:

PR #44, opened 2020-07-07
Add gitlab.com to the ssh-scan in the entrypoint

lluiscampos:


lluiscampos:
This should solve the last issue we had in the k8s deployed bot: it was failing on authenticating gitlab.com and therefore the pr_XXX branches not being created.

PR #49, opened 2020-09-01
QA-201: Vendor dependabot PR's

oleorhagen:
Since Github's Dependabot does not support vendoring in Go, a manual step has to
be applied after a PR is opened with a dependency upgrade.

This commit will add the functionality, by having the bot fetch the remote
branch, vendor and tidy the repository, and push the new commit.

Changelog: None

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -2,10 +2,13 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
 	"os/exec"
 	"strings"
 	"time"
 
+	"github.com/google/go-github/v28/github"
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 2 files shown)


helpers.go
```diff
@@ -70,3 +73,118 @@ func getListOfVersionedRepositories(inVersion string) ([]string, error) {
 
 	return strings.Split(strings.TrimSpace(string(output)), "\n"), nil
 }
+
+func validDependabotPR(pr *github.PullRequestEvent) bool {
+	// NOTE: For now, checking the 'labeled' condition, and not 'opened' (for testing purposes)
```
oleorhagen, line 78: Unsure, whether this is the right check, once this is live

helpers.go
```diff
@@ -70,3 +73,118 @@ func getListOfVersionedRepositories(inVersion string) ([]string, error) {
...
+		return fmt.Errorf("%v returned error: %s: %s", cmd.Args, out, err.Error())
+	}
+
+	log.Debug("Checking out the local branch..")
+	cmd = exec.Command("git", "checkout", "--quiet", localBranch)
+	cmd.Dir = tmpdir
+	out, err = cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("%v returned error: %s: %s", cmd.Args, out, err.Error())
+	}
```
lluiscampos, line 141: If `head` is just a regular branch name, we can simplify all these commands with just something like `"git clone -b "+head+ "git@github.com:mendersoftware/"+repo+".git"`

helpers.go
lluiscampos: I don't think this will be enough, as the bot might add different labels. I think it is safer with the "open" condition, but maybe you had a good reason for not using it?

oleorhagen: Yes, testing locally with the request you sent me :P 
But for the final, opened should be it. Just, could you please double check that this shows up in of the the `webhook` requests?

helpers.go
oleorhagen: I don't know what this magic is, but sure :smile_cat:

oleorhagen:
@lluiscampos It seems we can also get these issues, from time to time:
[pipeline](https://gitlab.com/Northern.tech/Mender/mender-artifact/-/pipelines/184290914)

helpers.go
oleorhagen: Should we also add the `--single-branch` option then, for clone economy?

lluiscampos: @oleorhagen Good idea

helpers.go
lluiscampos: Sorry. I sent you again a new one, this one should be correct.

lluiscampos:
> @lluiscampos It seems we can also get these issues, from time to time:
> [pipeline](https://gitlab.com/Northern.tech/Mender/mender-artifact/-/pipelines/184290914)

For now, I think manual intervention in these cases is fine... Let's get _something_ in place and improve later.

oleorhagen:
@lluiscampos this should be ready for final review now :)

helpers.go
```diff
@@ -70,3 +73,85 @@ func getListOfVersionedRepositories(inVersion string) ([]string, error) {
...
+	defer os.RemoveAll(tmpdir)
+
+	cmd := exec.Command("git", "clone", "--single-branch", "--branch", branchName, "git@github.com:mendersoftware/"+repo+".git")
+	cmd.Dir = tmpdir
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("%v returned error: %s: %s", cmd.Args, out, err.Error())
+	}
+
+	cmdDir := tmpdir + "/mender-artifact"
```
lluiscampos, line 110: `mender-artifact`? I don't understand...

lluiscampos: Ah, I see, you need `repo` here instead.

lluiscampos (changes requested):
See my comment above.

helpers.go
oleorhagen: lol, fixing

oleorhagen: Good thing we have more eyes on this

oleorhagen: thx

oleorhagen:
Also, another question to think about:
Do we need a git user set up? :thinking:

lluiscampos:
> Also, another question to think about:
> Do we need a git user set up? thinking

The Git user (and SSH credentials) are in place already.

lluiscampos (approved):
Nice! Let's merge :crossed_fingers:

oleorhagen:
Nice! One less thing to worry about. Let's give it a spin!

PR #54, opened 2020-09-11
Handle meta-mender PRs targeting "master-next"

lluiscampos:
By skipping any builds.

Also:
* Clean references to our old CI server
* Remove BASE_BRANCH parameter, which is unused in the pipeline
* Fix typos in comment

PR #57, opened 2020-09-29
Re-enable meta-mender builds on PR's

oleorhagen:
If I'm not mistaken, there is a slight logic error here: https://github.com/mendersoftware/integration-test-runner/pull/54/commits/f4b4cc504e53eba967e94a26c9ab5f375ccd2358

Second commit is optional. I just liked the sound better :)

### DIFF
--- a/main.go
+++ b/main.go
@@ -256,7 +256,7 @@ func main() {
 
 			for idx, build := range builds {
 				log.Infof("%d: "+spew.Sdump(build)+"\n", idx+1)
-				if build.repo == "meta-mender" && build.baseBranch != "master-next" {
+				if build.repo == "meta-mender" && build.baseBranch == "master-next" {
 					log.Info("Skipping build for meta-mender:master-next")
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -256,8 +256,8 @@ func main() {
 
 			for idx, build := range builds {
 				log.Infof("%d: "+spew.Sdump(build)+"\n", idx+1)
-				if build.repo == "meta-mender" && build.baseBranch != "master-next" {
-					log.Info("Skipping build for meta-mender:master-next")
+				if build.repo == "meta-mender" && build.baseBranch == "master-next" {
```
lluiscampos, line 259: Ouch! :sweat_smile:

lluiscampos (approved):
Thanks. It was obviously wrong.

main.go
oleorhagen: haha, I've done way worse

PR #61, opened 2020-11-03
Sync OS to Ent forks for feature branches too

lluiscampos:


main.go
```diff
@@ -674,14 +674,14 @@ func syncIfOSHasEnterpriseRepo(conf *config, gpr *github.PullRequestEvent) error
...
-		// verify release branches.
+		// Only sync on Merges to master, release and feature branches.
+		// Verify release branches.
 		branch := pr.GetBase()
 		if branch == nil {
 			return fmt.Errorf("syncIfOSHasEnterpriseRepo: Failed to get the base-branch of the PR: %v", branch)
 		}
 
-		syncBranches := regexp.MustCompile(`(master|[0-9]+\.[0-9]+\.x)`)
+		syncBranches := regexp.MustCompile(`(master|[0-9]+\.[0-9]+\.x|feature-.+)`)
```
oleorhagen, line 684: Question:
Just to double check here:
Is this the right syntax for `feature` branches?

oleorhagen (approved):

main.go
lluiscampos: @oleorhagen Yes, you can see it for example in [the pipeline](https://github.com/mendersoftware/mendertesting/blob/master/.gitlab-ci-check-docker-build.yml#L87).

(I'll slightly rework this PR to re-use a recently introduced constant for this)

oleorhagen (approved):
Looks correct to me.

I guess a raw string is not strictly necessary here though, and something like:

```
regexp.MustCompile("(master|...|%s.+)", featureBranchPrefix)
```
would be as nice

PR #63, opened 2020-11-26
Fix typo in conflict's comment

lluiscampos:
Parameter `repoOS` was not being set, so the resulting comment was always
something like `mender-test-bot/-enterprise.git`.

Fix by just remove this parameter and using `repo` instead.

### DIFF
--- a/main.go
+++ b/main.go
@@ -927,7 +927,6 @@ type commentArgs struct {
 	conf           *config
 	mergeConflicts bool
 	repo           string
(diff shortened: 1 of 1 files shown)


lluiscampos:
See for example https://github.com/mendersoftware/deployments-enterprise/pull/257#issuecomment-734161755

oleorhagen (approved):
Thanks

PR #65, opened 2021-01-06
Sync repos configurable

lluiscampos:
* Configurable list of repos for which to do GitHub->GitLab sync
Rework a bit the config to allow for a configurable list of repositories
to sync via env variable WATCH_REPOS_SYNC.

* Make Git config user/email configurable via env variables
These are only required for operations that perform "git commit", like
dependabot's commits amend.

* Add some minimal documentation on the three main features

* Extend the sync and PR branches features to CFEngine
Namely, map GitHub organizations to GitLab groups, so that we can use
these features on both organizations.

### DIFF
--- a/entrypoint
+++ b/entrypoint
@@ -1,5 +1,8 @@
 #!/bin/sh
 
+git config --global user.name ${GIT_CONFIG_USER_NAME-Mender Root}
+git config --global user.email ${GIT_CONFIG_USER_EMAIL-root@mender-jenkins.mender.io}
+
 ssh-keyscan github.com >> /root/.ssh/known_hosts
 ssh-keyscan gitlab.com >> /root/.ssh/known_hosts
 
--- a/main.go
+++ b/main.go
@@ -28,13 +28,13 @@ import (
 var mutex = &sync.Mutex{}
 
 type config struct {
-	githubSecret               []byte
-	githubToken                string
-	gitlabToken                string
-	gitlabBaseURL              string
-	watchRepositories          []string
-	integrationBranchDependant []string
-	integrationDirectory       string
+	githubSecret                     []byte
+	githubToken                      string
+	gitlabToken                      string
+	gitlabBaseURL                    string
+	integrationDirectory             string
+	watchRepositoriesTriggerPipeline []string // List of repositories for which to trigger mender-qa pipeline
+	watchRepositoriesGitLabSync      []string // List of repositories for which to trigger GitHub->Gitlab branches sync
 }
 
 type buildOptions struct {
@@ -45,11 +45,40 @@ type buildOptions struct {
 	makeQEMU   bool
 }
 
+// List of repos for which the integration pipeline shall be run
+// It can be overridden with env. variable WATCH_REPOS_PIPELINE
+var defaultWatchRepositoriesPipeline = []string{
+	"create-artifact-worker",
+	"deployments",
+	"deployments-enterprise",
+	"deviceadm",
+	"deviceauth",
+	"inventory",
+	"inventory-enterprise",
+	"integration",
+	"mender",
+	"mender-artifact",
+	"mender-conductor",
+	"mender-conductor-enterprise",
+	"meta-mender",
+	"mender-api-gateway-docker",
+	"tenantadm",
+	"useradm",
+	"useradm-enterprise",
+	"workflows",
+	"workflows-enterprise",
+	"auditlogs",
+	"mtls-ambassador",
+	"mender-connect",
+}
+
 var githubRepoToGitLabProject = map[string]string{
 	"saas": "Northern.tech/MenderSaaS/saas",
 }
 
-var enterpriseRepositories = []string{
+// List of repos for which the GitHub->Gitlab sync shall be performed.
+// It can be overridden with env. variable WATCH_REPOS_SYNC
+var defaultWatchRepositoriesSync = []string{
 	// backend
 	"deployments-enterprise",
 	"inventory-enterprise",
@@ -104,7 +133,8 @@ func initLogger() {
 }
 
 func getConfig() (*config, error) {
-	var repositoryWatchList []string
+	var repositoryWatchListPipeline []string
+	var repositoryWatchListSync []string
 	githubSecret := os.Getenv("GITHUB_SECRET")
 	githubToken := os.Getenv("GITHUB_TOKEN")
 	gitlabToken := os.Getenv("GITLAB_TOKEN")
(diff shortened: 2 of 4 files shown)


lluiscampos:
@oleorhagen I haven't really tested it... Please put a good eye on it ;)

main.go
```diff
@@ -45,11 +45,40 @@ type buildOptions struct {
 	makeQEMU   bool
 }
 
+// List of repos for which the integration pipeline shall be run
+// In can be overridden with env. variable WATCH_REPOS_PIPELINE
```
0lmi, line 49: Probably the typo here: In -> It

main.go
```diff
@@ -45,11 +45,40 @@ type buildOptions struct {
...
+	"mender-connect",
+}
+
 var githubRepoToGitLabProject = map[string]string{
 	"saas": "Northern.tech/MenderSaaS/saas",
 }
 
-var enterpriseRepositories = []string{
+// List of repos for which the GitHub->Gitlab sync shall be performed.
+// In can be overridden with env. variable WATCH_REPOS_SYNC
```
0lmi, line 80: Same here: In -> It

0lmi (commented):
@lluiscampos how do we specify github organisation? In the format `github.com/<organisation>/<repo>` for Mender it's `mendersoftware` but for CFEngine it should be `cfengine`.

lluiscampos:
> @lluiscampos how do we specify github organisation? In the format `github.com/<organisation>/<repo>` for Mender it's `mendersoftware` but for CFEngine it should be `cfengine`.

I realized that it is not necessary, the bot is organization agnostic. We deploy the bot somewhere, and then from `github.com/cfengine` settings we set a webhook to this deployment. So the bot only gets API calls from this organization, there is nothing else to set up.

(while typing the above) What we might need to configure is the GitLab organization... Let me look again into it.

lluiscampos:
> (while typing the above) What we might need to configure is the GitLab organization... Let me look again into it.

The last commit addresses this

main.go
```diff
@@ -45,11 +45,40 @@ type buildOptions struct {
 	makeQEMU   bool
 }
 
+// List of repos for which the integration pipeline shall be run
+// It can be overridden with env. variable WATCH_REPOS_PIPELINE
+var defaultWatchRepositoriesPipeline = []string{
```
oleorhagen, line 50: How about we just drop this, and only rely on the environment variable? :thinking: 
Do we even need this fallback?

main.go
```diff
@@ -124,40 +154,20 @@ func getConfig() (*config, error) {
...
-			"useradm-enterprise",
-			"workflows",
-			"workflows-enterprise",
-			"auditlogs",
-			"mtls-ambassador",
-			"mender-connect",
-		}
+	watchRepositoriesPipeline := os.Getenv("WATCH_REPOS_PIPELINE")
+
+	if len(watchRepositoriesPipeline) == 0 {
```
oleorhagen, line 166: Nit:
`watchRepositoriesPipeline == ""`
should be sufficient

main.go
```diff
@@ -326,7 +337,7 @@ func getBuilds(conf *config, pr *github.PullRequestEvent) []buildOptions {
 
 	makeQEMU := false
 
-	for _, watchRepo := range conf.watchRepositories {
+	for _, watchRepo := range conf.watchRepositoriesPipeline {
```
oleorhagen, line 348: Stretch, but maybe we should start thinking about splitting this up soon. In which case `config` would be a pretty self-contained package.

main.go
```diff
@@ -72,7 +72,14 @@ var defaultWatchRepositoriesPipeline = []string{
...
 
-var githubRepoToGitLabProject = map[string]string{
+// Mapping https://github.com/<org> -> https://gitlab.com/Northern.tech/<group>
+var gitHubOrganizationToGitLabGroup = map[string]string{
+	"mendersoftware": "Mender",
+	"cfengine":       "CFEngine",
+}
+
+// Mapping of special repos that have a custom group/project
+var gitHubRepoToGitLabProject = map[string]string{
```
oleorhagen, line 82: Suggestion:
`githubRepoToGitlabProjectSpecial` or something similar

main.go
```diff
@@ -501,11 +524,11 @@ func createPullRequestBranch(repo, pr, action string) error {
 		return fmt.Errorf("%v returned error: %s: %s", gitcmd.Args, out, err.Error())
 	}
 
-	// By default, assume the GitLab project is Northern.tech/Mender/<repo>
-	remoteURL := "git@gitlab.com:Northern.tech/Mender/" + repo
-	if v, ok := githubRepoToGitLabProject[repo]; ok {
-		remoteURL = "git@gitlab.com:" + v
+	remoteURL, err := getRemoteUrlGitLab(org, repo)
+	if err != nil {
+		return fmt.Errorf("getRemoteUrlGitLab returned error: %s", err.Error())
```
oleorhagen, line 529: Nit:
`err.Error()` isn't really necessary. The formatter will call `Error`

oleorhagen (approved):
Really really nice my friend!
I've tried hard spotting errors, but didn't find any (doesn't mean I'm right :-1: ).

See the comments, and ignore and pick at your will, mostly it's cosmetics, but I like cosmetics.

Todays first
:medal_sports:

entrypoint
```diff
@@ -1,5 +1,8 @@
 #!/bin/sh
 
+git config --global user.name ${GIT_CONFIG_USER_NAME-Mender Root}
```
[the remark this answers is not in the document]
lluiscampos: Very confusing phrase... We use this in many places in our scripts, so I'll leave it

main.go
lluiscampos: Yes, we are not setting the env variable in our Mender deployment, so we need the fallback.

We could update the env variable every time we have a new repo, but I find the code is easier to track for us developers than to go into an env variable in the SRE side of things.

main.go
lluiscampos: Maybe next time ;)

main.go
lluiscampos: I noticed here and earlier that you use "github" or "Gitlab" instead of "gitHub" or "GitLab". Do you have an strong opinion on this?

wrt the suggestion itself: okey.

main.go
lluiscampos: This is everywhere, so I'll rather leave as-is.

main.go
oleorhagen: Makes sense

main.go
oleorhagen: No, not at all. I just haven't realized till now that they actually write `GitHub` and `GitLab`

lluiscampos:
> I've tried hard spotting errors, but didn't find any (doesn't mean I'm right -1 ).

@oleorhagen It turns out that it doesn't work. Amend in #67

PR #67, opened 2021-01-06
Take the organization name from the top level Organization

lluiscampos:
The previous filed seems to not exist in pull_request events.

### DIFF
--- a/main.go
+++ b/main.go
@@ -264,7 +264,7 @@ func main() {
 
 			// To run component's Pipeline create a branch in GitLab, regardless of the PR
 			// coming from a mendersoftware member or not (equivalent to the old Travis tests)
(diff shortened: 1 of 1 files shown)


lluiscampos:
This PR changes handling on `pull_request` events to get the org via `organization.login`. For `push` events we leave it as is via `repository.organization`

According to the documentation we could use `organization.login` in both cases:
* https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#pull_request
* https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#push

But the go library we are using does not support it for `push` events (it might be fixed with an update, I haven't checked).

In any case I have verified the latest webhooks sends from GitHub to our bot and _I think_ it should work now for both cases.

oleorhagen (approved):

PR #69, opened 2021-01-22
Allow for user setting empty watch lists

lluiscampos:
Use os.LookupEnv to check for empty env variables, so that a user can
set WATCH_REPOS_PIPELINE or WATCH_REPOS_SYNC to "" to indicate no
repository in the list.

### DIFF
--- a/main.go
+++ b/main.go
@@ -162,20 +162,20 @@ func getConfig() (*config, error) {
 		}
 	}
 
-	watchRepositoriesTriggerPipeline := os.Getenv("WATCH_REPOS_PIPELINE")
+	watchRepositoriesTriggerPipeline, ok := os.LookupEnv("WATCH_REPOS_PIPELINE")
 
-	if watchRepositoriesTriggerPipeline == "" {
-		repositoryWatchListPipeline = defaultWatchRepositoriesPipeline
-	} else {
+	if ok {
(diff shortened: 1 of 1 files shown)


0lmi (approved):
LGTM 👍

alfrunes (approved):

main.go
```diff
@@ -162,20 +162,20 @@ func getConfig() (*config, error) {
...
 	}
 
-	watchRepositoriesGitLabSync := os.Getenv("WATCH_REPOS_SYNC")
+	watchRepositoriesGitLabSync, ok := os.LookupEnv("WATCH_REPOS_SYNC")
 
-	if watchRepositoriesGitLabSync == "" {
-		repositoryWatchListSync = defaultWatchRepositoriesSync
-	} else {
+	if ok {
 		repositoryWatchListSync = strings.Split(watchRepositoriesGitLabSync, ",")
```
merlin-northern, line 176: note that it still can be empty, set, but empty.

merlin-northern (approved):
lgtm assuming that when set it will not be set to empty string.

main.go
lluiscampos: @merlin-northern That is the exact use case that I am trying to support with this PR

lluiscampos:
> lgtm assuming that when set it will not be set to empty string.

@merlin-northern But that is exactly its purpose! To let users set it to empty string!

PR #71, opened 2021-01-28
Move code into different files + Add log field with unique ID + Process webhooks on Goroutine

lluiscampos:
It looks like a lot, but it should be fairly easy to follow commit by commit.


oleorhagen (approved):
Great, I don't really have anything to add.

I can actually navigate the repository now!

Only thing which came to mind was merging `pullrequest_pipeline` with `pullrequest_pipeline_helpers`, but yeh.. 

Nice split, cleaning was way quicker and easier than I feared!

lluiscampos:
> Only thing which came to mind was merging `pullrequest_pipeline` with `pullrequest_pipeline_helpers`, but yeh..

I thought about this too, and ironically decided to leave them separate for "historical reasons". It is actually the ONLY detail after this PR that keeps some sort of history :)

tranchitella (approved):
Looks good! The DeliverID tag will make log analysis way easier, well done.

PR #72, opened 2021-01-29
Correctly implement the delivery field in the logs

lluiscampos:
Using logrus Hook interface. Commit fdefa68 was insufficient.

### DIFF
--- a/log.go
+++ b/log.go
@@ -1,0 +1,40 @@
+package main
+
+import (
+	"os"
+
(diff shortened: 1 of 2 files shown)


lluiscampos:
@tranchitella The previous PR was really untested (and turned out wrong :) For this implementation I have done some dummy tests locally and it should work.

tranchitella (approved):

PR #73, opened 2021-02-01
Create logger from context and pass it around

lluiscampos:
So that each Goroutine has its own logrus.Entry with the unique ID.

The previous implementation with the hook was not enough, as the field
was being overridden globally on every new POST.


lluiscampos:
@alfrunes  I think you might be unfamiliar to this code base but I am confident that you can provide me very good feedback! I was having a problem with concurrency and the global `log` object, that I tried to fix by setting/getting this logger from the `gin.Context`. Please have a look :)

main.go
```diff
@@ -293,12 +317,11 @@ func main() {
...
-			log.Warnln("payload failed to validate, ignoring.")
+			logrus.Warnln("payload failed to validate, ignoring.")
 			return
 		}
 
-		// Set unique ID for the logger
-		log.AddHook(newDeliveryFieldHook(github.DeliveryID(context.Request)))
+		context.Set("delivery", github.DeliveryID(context.Request))
 
 		go processGitHubWebhook(context, payload, githubClient, conf)
```
alfrunes, line 326: The request handler is already running in a separate go routine.
```suggestion
		processGitHubWebhook(context, payload, githubClient, conf)
```

alfrunes (approved):
This repository is indeed structured a bit different than I am used to.
The PR looks OK, though I find it a bit strange passing the `logrus.Entry` around.
Personally, I would prefer having the handler (or a simple middleware) setup the request parameters in a `context.Context` and have a `LoggerFromContext` interface to instantiate a new `logrus.Entry` where needed.
I left a couple of comments (mostly outside the diff) with some suggestions you may take as you wish.

main.go
lluiscampos: Maybe, but what we are trying to achieve here is to reply to GitHub before the full processing of the request.

This was actually a big problem before, where many of the requests were timing out (from GitHub point of view) and we could see in `gin` logs that we only responded after all the processing was done.

main.go
```diff
@@ -195,6 +217,8 @@ func processGitHubWebhook(context *gin.Context, payload []byte, githubClient *gi
 	webhookType := github.WebHookType(context.Request)
 	webhookEvent, _ := github.ParseWebHook(github.WebHookType(context.Request), payload)
 
+	log := getCustomLoggerFromContext(context)
```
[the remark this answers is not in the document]
lluiscampos: Noted, thanks

lluiscampos:
> This repository is indeed structured a bit different than I am used to.
> The PR looks OK, though I find it a bit strange passing the `logrus.Entry` around.
> Personally, I would prefer having the handler (or a simple middleware) setup the request parameters in a `context.Context` and have a `LoggerFromContext` interface to instantiate a new `logrus.Entry` where needed.

I was browsing code in backend services and `go-lib-micro` where the design is a bit more as you describe. In the one hand it looked a bit of an overkill for this repo, and on the other hand I realized that we would only use the context to intansiate a `logrus.Entry` virtually everywhere, so I thought that passing the log would be more efficient and same ugly/nice.

Anyway, I'll keep it in the back of my head, maybe this could become a proper backend quality repo at some point!

> I left a couple of comments (mostly outside the diff) with some suggestions you may take as you wish.

Thanks for your review!

main.go
alfrunes: Ahh, OK. Makes sense.

PR #74, opened 2021-02-11
QA-264: Delete pr_x branch in Gitlab upon a close or merge

oleorhagen:
Changelog: None

### DIFF
--- a/main.go
+++ b/main.go
@@ -245,6 +245,11 @@ func processGitHubWebhook(ctx *gin.Context, payload []byte, githubClient *github
 			log.Errorf("Could not create PR branch: %s", err.Error())
 		}
 
+		// Delete merged pr branches in GitLab
+		if err = deleteStaleGitlabPRBranch(log, pr, conf); err != nil {
+			log.Errorf("Failed to delete the stale PR branch after the PR: %v was merged or closed. Error: %v", pr, err)
+		}
+
 		// Continue to the integration Pipeline only for mendersoftware members
 		if member, _, _ := githubClient.Organizations.IsMember(ctx, "mendersoftware", pr.Sender.GetLogin()); !member {
 			log.Warnf("%s is making a pullrequest, but he/she is not a member of mendersoftware, ignoring", pr.Sender.GetLogin())
--- a/pullreq_pipeline.go
+++ b/pullreq_pipeline.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 
+	"github.com/google/go-github/v28/github"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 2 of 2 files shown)


menderqa_pipeline.go
```diff
@@ -351,3 +355,57 @@ func stopBuildsOfStalePRs(log *logrus.Entry, pr *github.PullRequestEvent, conf *
 	return nil
 
 }
+
+func deleteStaleGitlabPRBranch(log *logrus.Entry, pr *github.PullRequestEvent, conf *config) error {
```
lluiscampos, line 359: This has nothing to do with `menderqa_pipeline`, it should be moved to `pullreq_pipeline`

main.go
```diff
@@ -263,6 +263,11 @@ func processGitHubWebhook(ctx *gin.Context, payload []byte, githubClient *github
 			log.Errorf("Failed to stop a stale build after the PR: %v was merged or closed. Error: %v", pr, err)
 		}
 
+		// Delete merged pr branches in GitLab
+		if err = deleteStaleGitlabPRBranch(log, pr, conf); err != nil {
+			log.Errorf("Failed to delete the stale PR branch after the PR: %v was merged or closed. Error: %v", pr, err)
+		}
+
```
lluiscampos, line 270: Similar to my previous comment, this should be done before line 248, as from there on the processing only continues for `mendersoftware` members (iow mender-qa pipeline)

menderqa_pipeline.go
```diff
@@ -351,3 +355,57 @@ func stopBuildsOfStalePRs(log *logrus.Entry, pr *github.PullRequestEvent, conf *
...
+
+	// If the action is "closed" the pull request was merged or just closed,
+	// stop builds in both cases.
+	if pr.GetAction() != "closed" {
+		log.Debugf("stopBuildsOfStalePRs: PR not closed, therefore not stopping it's pipeline")
```
lluiscampos, line 364: Typo in the function name

lluiscampos (changes requested):
Looks good otherwise.

pullreq_pipeline.go
```diff
@@ -68,3 +69,57 @@ func createPullRequestBranch(log *logrus.Entry, org, repo, pr, action string) er
 	log.Info("Pipeline is expected to start automatically")
 	return nil
 }
+
+func deleteStaleGitlabPRBranch(log *logrus.Entry, pr *github.PullRequestEvent, conf *config) error {
```
lluiscampos, line 73: For consistency, consider the same interface as the function above

```suggestion
func deleteStaleGitlabPRBranch(log *logrus.Entry, org, repo, pr, action string) error {
```

lluiscampos (approved):
Minor suggestion, change only if you wish.

pullreq_pipeline.go
oleorhagen: I agreed, looking at it, but from looking in the main code, I actually think this is the saner interface, so leaving for now :)

PR #75, opened 2021-02-12
QA-255: Do not build the mender-client from source unneeded

oleorhagen:
This takes advantage of the functionality in:

https://github.com/mendersoftware/mender-qa/pull/481

And sets BUILD_CLIENT == false, in the case a repository in the
prebuiltClientRepositories array is targeted.

Changelog: None

### DIFF
--- a/menderqa_pipeline.go
+++ b/menderqa_pipeline.go
@@ -315,6 +315,16 @@ func getBuildParameters(log *logrus.Entry, conf *config, build *buildOptions) ([
 	buildParameters = append(buildParameters, &gitlab.PipelineVariable{Key: "TEST_VEXPRESS_QEMU_UBOOT_UEFI_GRUB", Value: qemuParam})
 
 	buildParameters = append(buildParameters, &gitlab.PipelineVariable{Key: "BUILD_BEAGLEBONEBLACK", Value: qemuParam})
+
+	// Set BUILD_CLIENT = false, if target repo not in the qemuBuildRepositories list
+	buildClient := &gitlab.PipelineVariable{Key: "BUILD_CLIENT", Value: "false"}
+	for _, prebuiltClientRepo := range qemuBuildRepositories {
+		if build.repo == prebuiltClientRepo {
+			buildClient.Value = "true"
+		}
+	}
+	buildParameters = append(buildParameters, buildClient)
+
 	return buildParameters, nil
 }
 


oleorhagen:
This goes along with: https://github.com/mendersoftware/mender-qa/pull/481

kacf (approved):

main.go
```diff
@@ -103,6 +103,35 @@ var defaultWatchRepositoriesSync = []string{
 	"mender-qa",
 }
 
+// List of repositories which runs integration tests, but do not need a fresh
+// built client
+var prebuiltClientRepositories = []string{
```
lluiscampos, line 108: I think having a new list of repos is not wise: adding any repository we need to modify both `defaultWatchRepositoriesPipeline` and this one.

main.go
```diff
@@ -103,6 +103,35 @@ var defaultWatchRepositoriesSync = []string{
...
+	"create-artifact-worker",
+	"deployments",
+	"deployments-enterprise",
+	"deviceadm",
+	"deviceauth",
+	"inventory",
+	"inventory-enterprise",
+	"integration",
+	// "mender",
+	"mender-artifact",
```
lluiscampos, line 118: I guess this is the reason for the new list. I would argue that `mender-artifact` cannot use a pre-compiled yocto image -> mender-artifact is part of the image.

In my understanding, only repos not in `qemuBuildRepositories` are allowed to use a pre-compiled yocto image. (this links with my previous comment, as I don't think we need two lists)

menderqa_pipeline.go
```diff
@@ -315,6 +315,14 @@ func getBuildParameters(log *logrus.Entry, conf *config, build *buildOptions) ([
 	buildParameters = append(buildParameters, &gitlab.PipelineVariable{Key: "TEST_VEXPRESS_QEMU_UBOOT_UEFI_GRUB", Value: qemuParam})
 
 	buildParameters = append(buildParameters, &gitlab.PipelineVariable{Key: "BUILD_BEAGLEBONEBLACK", Value: qemuParam})
+
+	// Set BUILD_CLIENT == false, if target repo in prebuiltClientRepositories list
```
lluiscampos, line 319: nitpick
```suggestion
	// Set BUILD_CLIENT = false, if target repo in prebuiltClientRepositories list
```

lluiscampos (changes requested):
See my comment below. Does it make sense to you? Can you justify `mender-artifact` using a pre-build image?

lluiscampos (approved):
Thanks! :100:

lluiscampos:
(Reminder: we cannot merge this PR before https://github.com/mendersoftware/mender-qa/pull/481)

PR #76, opened 2021-02-16
QA-232: suggest cherry-picks to the latest three release branches

tranchitella:


tranchitella:
@lluiscampos any suggestion on how to get rid of the `Host key verification failed` (see the failed job in the CI pipeline)?

tranchitella:
If we want to run unit tests, I need the goveralls setup. Before proceeding, though, I'm waiting for your feedbacks @lluiscampos and @oleorhagen

menderqa_pipeline_helpers.go
```diff
@@ -9,8 +9,21 @@ import (
...
+	if err != nil {
+		err = fmt.Errorf("getLatestIntegrationRelease: Error: %v (%s)", err, version)
+	}
+	versionStr := strings.TrimSpace(string(version))
+	return strings.SplitN(versionStr, "\n", -1), err
+}
+
 func getServiceRevisionFromIntegration(repo, baseBranch string, conf *config) (string, error) {
-	c := exec.Command("release_tool.py", "--version-of", repo, "--in-integration-version", baseBranch)
+	c := exec.Command("python3", "release_tool.py", "--version-of", repo, "--in-integration-version", baseBranch)
```
oleorhagen, line 26: Question:
Why was this needed?

oleorhagen (approved):
I think it looks good.

Prior to reviewing this I was thinking along the lines, that maybe we should have the ability to modify the predicate it matches on for notifying about cherries needed (like a function which takes the commit message (string), and returns true of false, and is then mapped onto the commit messages in order). However, it's not necessary. 

Also, prior to reviewing I wasn't sure if `Changelogs` are the best measure here. But you guys are probably right (I was thinking along the lines of having something like `fix` or `fixes` in the commit message would trigger this.).

cherrypicks.go
```diff
@@ -0,0 +1,142 @@
...
+	gitcmd = exec.Command("git", "fetch", "github", "pull/"+prNumber+"/head:"+prBranchName)
+	gitcmd.Dir = tmpdir
+	out, err = gitcmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("%v returned error: %s: %s", gitcmd.Args, out, err.Error())
+	}
+
+	// count the number of non-none changelog entries
+	baseSHA := *pr.PullRequest.Base.SHA
+	countCmd := exec.Command("sh", "-c", "git log "+baseSHA+"...pr_"+prNumber+" | grep -i \"Changelog: \" | grep -v -i \": none\" | wc -l")
```
kacf, line 79: Suggest being a bit more explicit, since it's not uncommon to include the word "changelog" in various other contexts for example when doing releases, and to have the word "none" in an actual changelog message:

```
grep -i \"^Changelog:\" | grep -v -i \"^Changelog: *none\"
```

Might be useful to have a test for these sentences too. Examples would be:

* ```
  Fix changelog: This was actually reverted.

  Cancel-changelog: abcdef0123456789
   ```
* ```
  Check for nil.

  Changelog: Fix API returning none.
```

File: deb-requirements.txt [hunk 3]
```diff
@@ -0,0 +1 @@
+python3-yaml
```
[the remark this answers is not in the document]
kacf: Not if the dependency on this component is introduced in this commit.

menderqa_pipeline_helpers.go
```diff
@@ -9,8 +9,21 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+func getLatestIntegrationRelease(number int, conf *config) ([]string, error) {
+	cmd := fmt.Sprintf("git for-each-ref --sort=-creatordate --format='%%(refname:short)' 'refs/tags' "+
+		"| sed -E '/(^[0-9]+\\.[0-9]+)\\.[0-9]+$/!d;s//\\1.x/' | uniq | head -n %d | sort -n -r", number)
```
kacf, line 14: Suggestion: Instead of `sort -n`, use `sort -V`, which is specifically for sorting versions.

deb-requirements.txt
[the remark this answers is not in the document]
tranchitella: I need it to run unit tests, why do I need to separate it in a different commit?

entfork_sync.go
```diff
@@ -14,8 +14,9 @@ import (
 	"text/template"
 
 	"github.com/google/go-github/v28/github"
-
 	"github.com/sirupsen/logrus"
+
+	tgithub "github.com/mendersoftware/integration-test-runner/client/github"
```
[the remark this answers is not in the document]
tranchitella: because "github" is the import of the GitHub library, I need to use an alias because it is clashing with my github client, which I introduced to be able to mock it in unit tests.

main_test.go
```diff
@@ -0,0 +1,23 @@
...
+	cmd := exec.Command("sh", "-c", "! [ -f $HOME/.ssh/known_hosts ] && "+
+		"mkdir -p $HOME/.ssh && ssh-keyscan github.com | "+
+		"tee -a $HOME/.ssh/known_hosts && chmod 600 $HOME/.ssh/known_hosts || true")
+	err := cmd.Run()
+	if err != nil {
+		panic(err)
+	}
+	// import the SSH private key
+	cmd = exec.Command("sh", "-c", "[ -n \"$SSH_PRIVATE_KEY\" ] && ! [ -f $HOME/.ssh/id_rsa ] && "+
+		"echo \"$SSH_PRIVATE_KEY\" | tr -d '\r' | tee $HOME/.ssh/id_rsa && chmod 600 $HOME/.ssh/id_rsa || true")
```
kacf, line 18: We should not use `tee` for a private key, suggest using straight redirection so it doesn't show up in the log.

menderqa_pipeline_helpers.go
tranchitella: Because it doesn't work if you don't have "." in your PATH (like on my Mac OS).
Unit tests will fail because release_tool.py is not executable directly. It's either "./release_tool.py" or "python3 release_tool.py".

deb-requirements.txt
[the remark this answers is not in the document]
oleorhagen: I see, I figured it was missing from old.

entfork_sync.go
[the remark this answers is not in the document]
oleorhagen: Yes, but `t` for `temp`? :thinking:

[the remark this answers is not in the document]
tranchitella: t for `test-runner`, that was my intention :)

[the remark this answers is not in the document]
tranchitella: let's rename it to client_github

README.md
```diff
@@ -21,35 +21,29 @@ For all repositories in the organization, a pr_XXX branch will be created in
 GitLab for every pull/XXX PR from GitHub.
```
lluiscampos, line 22: Thanks for updating README

cherrypicks.go
```diff
@@ -0,0 +1,142 @@
...
+	countCmd.Dir = tmpdir
+	out, err = countCmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("%v returned error: %s: %s", gitcmd.Args, out, err.Error())
+	}
+
+	changelogs, _ := strconv.Atoi(strings.TrimSpace(string(out)))
```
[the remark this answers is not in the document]
tranchitella: TBH, I like strconv.Atoi. Can you elaborate more on the advantages of binary/encoding?

cherrypicks.go
tranchitella: Nice, thanks! Your grep machinery is way better than mine, I applied it.

menderqa_pipeline_helpers.go
```diff
@@ -9,8 +9,21 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+func getLatestIntegrationRelease(number int, conf *config) ([]string, error) {
```
lluiscampos, line 12: This function does not belong here: `menderqa_pipeline_helpers.go` are meant to be used only from `menderqa_pipeline.go`. Maybe you want to put it in `util.go`, or directly in `cherrypicks.go` (it is not used anywhere else).

main.go
```diff
@@ -264,14 +251,19 @@ func processGitHubWebhook(ctx *gin.Context, payload []byte, githubClient *github
...
 		// Keep the OS and Enterprise repos in sync
-		if err = syncIfOSHasEnterpriseRepo(log, conf, pr); err != nil {
+		if err := syncIfOSHasEnterpriseRepo(log, conf, pr); err != nil {
 			log.Errorf("Failed to sync the OS and Enterprise repos: %s", err.Error())
 		}
+
+		// If the pr was merged, suggest cherry-picks
+		if err := suggestCherryPicks(log, pr, githubClient, conf); err != nil {
+			log.Errorf("Failed to suggest cherry picks for the pr %v. Error: %v", pr, err)
+		}
```
lluiscampos, line 266: This has to be done above line 240, or else won't be processed for PRs from contributors.

main_test.go
```diff
@@ -0,0 +1,23 @@
...
+		"mkdir -p $HOME/.ssh && ssh-keyscan github.com | "+
+		"tee -a $HOME/.ssh/known_hosts && chmod 600 $HOME/.ssh/known_hosts || true")
+	err := cmd.Run()
+	if err != nil {
+		panic(err)
+	}
+	// import the SSH private key
+	cmd = exec.Command("sh", "-c", "[ -n \"$SSH_PRIVATE_KEY\" ] && ! [ -f $HOME/.ssh/id_rsa ] && "+
+		"echo \"$SSH_PRIVATE_KEY\" | tr -d '\r' >> $HOME/.ssh/id_rsa && chmod 600 $HOME/.ssh/id_rsa || true")
+	err = cmd.Run()
```
lluiscampos, line 19: I don't understand what is the intention here. We are not going to set an SSH_PRIVATE_KEY in CI variables, so is only meant for local use? Can the tests run without the SSH key at all?

lluiscampos (changes requested):
Having already @kacf and @oleorhagen eyes on the code, I focused more on high level things.

See my comments below; the only one "requesting for changes" is the one in `main.go`, as I believe we need to move this check earlier in the process.

Thank you very much for introducing unit tests. This is going to help everyone working on the tool :heart:

cherrypicks.go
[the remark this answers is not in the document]
oleorhagen: You just don't have to go via strings, but as mentioned: `feel free to choose`. And it seems you have :)

main_test.go
tranchitella: Without the SSH key, the unit tests cannot clone the repos, thus unit tests won't run in CI. Am I wrong?

menderqa_pipeline_helpers.go
tranchitella: moved

main.go
tranchitella: right, done

main_test.go
lluiscampos: Exactly, but we cannot have an *unprotected* SSH key on CI variables. I saw we have it right now and I assumed you added it while developing, but we definitely have to remove before merge, so the tests won't be able to run.

tranchitella: ok, which options do I have, then? maybe we can discuss it on slack.

tranchitella: reworked to work with https links when running unit tests

tranchitella:
@lluiscampos I removed the SSH private key variable in GitLab and implemented here (https://github.com/mendersoftware/integration-test-runner/pull/76/commits/be199cb754733167d53125f30157375411f9e34c) the possibility to use HTTPS urls in unit tests.
Can you please review the PR again?

util_test.go
```diff
@@ -0,0 +1,15 @@
...
+	"gotest.tools/assert"
+)
+
+func TestGetRemoteURLGitHub(t *testing.T) {
+	url := getRemoteURLGitHub(GitProtocolSSH, "mendersoftware", "workflows")
+	assert.Equal(t, "git@github.com:/mendersoftware/workflows.git", url)
+
+	url = getRemoteURLGitHub(GitProtocolHTTP, "mendersoftware", "workflows")
+	assert.Equal(t, "https://github.com/mendersoftware/workflows", url)
+}
```
lluiscampos, line 15: :+1:

lluiscampos (approved):
@tranchitella Excellent work. Let's deploy it!

PR #79, opened 2021-02-21
QA-254: start integration pipelines on demand using a command

tranchitella:


tranchitella:
@lluiscampos @oleorhagen comments, feedback?

main_comment.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// process created actions only, ignore the others
+	action := comment.GetAction()
+	if action != "created" {
+		log.Infof("Ignoring action %s on comment", action)
+		return nil
+	}
+
+	// accept commands only from organization members
+	if member := githubClient.IsOrganizationMember(ctx, githubOrganization, comment.Sender.GetLogin()); !member {
```
oleorhagen, line 25: Nit:
`member is stored as a bool and never used`

main_comment.go
```diff
@@ -0,0 +1,94 @@
...
+		log.Warnf("%s commented, but he/she is not a member of our organization, ignoring", comment.Sender.GetLogin())
+		return nil
+	}
+
+	// filter comments mentioning the bot
+	body := comment.Comment.GetBody()
+	if !strings.Contains(body, "@"+githubBotName) {
+		log.Infof("ignoring comment not mentioning me")
```
oleorhagen, line 33: Nit:
Format specifier, but no format string

main_comment.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil
+	}
+
+	// extract the command and check it is valid
+	command := ""
+	if strings.Contains(body, commandStartPipeline) {
```
oleorhagen, line 39: Question:
Could not all of this be a part of the switch statement itself? :thinking:

main_pullrequest.go
```diff
@@ -0,0 +1,90 @@
...
+	if pr.GetPullRequest().GetDraft() {
+		log.Infof("The PR: %s/%d is a draft. Do not run tests", pr.GetRepo().GetName(), pr.GetNumber())
+		return nil
+	}
+
+	if isDependabotPR, err := maybeVendorDependabotPR(log, pr, conf); isDependabotPR {
+		if err != nil {
+			log.Errorf("maybeVendorDependabotPR: %v", err)
+		}
+		return err
```
oleorhagen, line 25: suggestion:
Just be explicit and return `nil`

main_pullrequest.go
```diff
@@ -0,0 +1,90 @@
...
+func processGitHubPullRequest(ctx *gin.Context, pr *github.PullRequestEvent, githubClient clientgithub.Client, conf *config) error {
+	log := getCustomLoggerFromContext(ctx)
+
+	// Do not run if the PR is a draft
+	if pr.GetPullRequest().GetDraft() {
+		log.Infof("The PR: %s/%d is a draft. Do not run tests", pr.GetRepo().GetName(), pr.GetNumber())
+		return nil
+	}
+
+	if isDependabotPR, err := maybeVendorDependabotPR(log, pr, conf); isDependabotPR {
```
oleorhagen, line 21: Question:
Is the `isDependabotPR` check necessary. Could it not be more like a:
```
if _, err := maybeVendorDependabotPR(log, pr, conf); err != nil {
...
```
it seems more like no-op in the no-err case.

oleorhagen (approved):
Real nice!
The functionality in itself looks real nice. The refactoring makes sense, and the added tests are long overdue - thanks!

main_comment.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	switch command {
+	case commandStartPipeline:
+		// retrieve the pull request
+		prLink := comment.Issue.GetPullRequestLinks().GetURL()
+		if prLink == "" {
+			log.Warnf("ignoring comment not on a pull request")
+			return nil
+		}
```
lluiscampos, line 54: Shouldn't we do this at the beginning? Out of the switch case, at least

main_comment.go
```diff
@@ -0,0 +1,94 @@
...
+			return err
+		}
+
+		// make sure we only parse one pr at a time, since we use release_tool
+		mutex.Lock()
+
+		// get the list of builds
+		prRequest := &github.PullRequestEvent{Repo: comment.GetRepo(), PullRequest: pr}
+		builds := parsePullRequest(log, conf, "opened", prRequest)
+		log.Infof("%s:%d would trigger %d builds", comment.GetRepo().GetName(), pr.GetNumber(), len(builds))
```
lluiscampos, line 75: Suggestion: `will trigger` or `is triggering`.

lluiscampos (approved):
I can't add anything from what Ole said... Hats off to you :tophat:

main_comment.go
tranchitella: fixed

main_comment.go
tranchitella: fixed

main_pullrequest.go
tranchitella: fixed

main_pullrequest.go
tranchitella: I don't think so; if isDependabotPR is true, we need to stop there; if we get an error, we log it and stop here
I don't think isDependabotPR can be true in case of error, so I adjusted a bit the if condition.
Does it make sense?

main_comment.go
tranchitella: Done

main_comment.go
tranchitella: Moved outside the switch

main_comment.go
tranchitella: done

PR #87, opened 2021-04-27
Only comment once on a PR

oleorhagen:
The bot can be kinda noisy if there are a lot of pushes to a PR.

This makes sure that if there is a comment from the bot already, there is no new
comment posted.

Changelog: None


oleorhagen:
Ref: https://northern-tech.slack.com/archives/C0XM0KX9C/p1619514145327300

Untested, but so is the rest of the code in this file. In the matter of an error however, this should default back to the old behaviour, as far as I can tell, which is commenting always.

tranchitella:
@oleorhagen tests are failing :)

tranchitella (approved):

PR #89, opened 2021-05-05
[pullreq_pipeline] Do not create branches for PRs coming from branches

lluiscampos:
The functionality is meant to do branches from our forks, but some times
we do have PRs coming from actual branches, ending up with duplicate
pipelines running...

### DIFF
--- a/main_pullrequest.go
+++ b/main_pullrequest.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"strconv"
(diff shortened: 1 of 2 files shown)


lluiscampos:
It should fix these kind of double pipelines:
Screenshot_2021-05-05_16-04-04

oleorhagen (approved):
Looks good :)

Thanks for the cleanup

PR #92, opened 2021-05-07
RFC: Add functionality for automatically cherry-picks to release branches

oleorhagen:
With this functionality the bot will check the tags given to a PR, and if the
tag matches an existing release-branch. Thus if there is a match, the bot will
create a PR with the given cherry-pick (given no merge conflicts).

This is still clearly not working, I just want to see if you think it worthwhile to finish this?


oleorhagen:
> I'm a bit unsure. On one hand, it sounds practical, but on the other hand, it's quite often that the decision to cherry-pick to a release branch is wrong, for various reasons (for example you added a feature, but the bot thinks it's a bug fix that needs to be cherry-picked). I'm wondering if this falls into the same category as the automatic mender-qa build runner. It created too many builds and we ended up making it opt-in instead.
> 
> Could it be comment-based? And even better, could you post which versions you want in the comment? :-)

Sure, let me just make sure that we are on the same page here.

The idea is that it will only cherry-pick to branches given as labels in a PR, so:

We will create labels matching the release branches, and then adding the correct labels will be part of the review process, and in fact, a PR is not ready to be merged until the correct release branch [labels](https://docs.github.com/en/github/managing-your-work-on-github/managing-labels) are applied to the PR.

In this way, getting the correct picks will be both a part of the review process, and easily tracked.

This is fully transparent, and not automatic (the labels will have to be added manually), and no labels = no cherries.

Are we on the same page here?

kacf:
Aha! I completely missed this part. When you said "tags" I thought you meant Git tags, IOW the existing approach.

Yes, that sounds perfect, using the PR tags is a great idea!

lluiscampos (commented):
While I think using the PR labels for this is a cool idea, I think we should have *one* interface with the bot and be consistent.

At the moment we have a "comments interface", where we can send commands (currently only one). So with that in mind, I suggest we design this in way that we can use such commands.

For example: `mender-test-bot cherry-pick 2.7.x 2.6.x` could be the command to trigger the actual cherry-picks. In this regard, the current message that says "hei, consider if this PR needs cherrypicking" could give you the example commands of what exactly can you type to get it running (or, as some of us like to do, just reply to that message).

Does it make sense? My fear is that we will end-up with cool features but little intuitive.

In any case I appreciate the initiative, cool stuff!

oleorhagen:
> While I think using the PR labels for this is a cool idea, I think we should have _one_ interface with the bot and be consistent.
> 

I see what you mean, and I kinda agree. Although, since I'm a little invested here, _I do really like the labels_ :p 

I think I'll just have an ask in `dev-mender`, and see what people think

> At the moment we have a "comments interface", where we can send commands (currently only one). So with that in mind, I suggest we design this in way that we can use such commands.
> 
> For example: `mender-test-bot cherry-pick 2.7.x 2.6.x` could be the command to trigger the actual cherry-picks. In this regard, the current message that says "hei, consider if this PR needs cherrypicking" could give you the example commands of what exactly can you type to get it running (or, as some of us like to do, just reply to that message).
> 
> Does it make sense? My fear is that we will end-up with cool features but little intuitive.

Makes perfect sense, if we go down this road, this will be the way, for sure.

kacf:
Ah, @lluiscampos has a good point. Although my gut reaction was that tags look really nice, I think the comments fit the "command oriented nature" better. For example, with tags, how do you retry a cherry-pick (say, after something has been merged on the release branch which unblocks a conflict)? Comments work better in this regard.

oleorhagen:
@lluiscampos see if you like the new interface :)

cherrypicks.go
```diff
@@ -135,3 +141,136 @@ Hello :smile_cat: This PR contains changelog entries. Please, verify the need of
...
+	commentText := `Hi :smileycat:
+I did my very best, and this is the result of the cherry pick operation:
+`
+	for _, targetBranch := range targetBranches {
+		if !conflicts[targetBranch] {
+			commentText = commentText +
+				fmt.Sprintf("\t* %s:white_check_mark:\n", targetBranch)
+		} else {
+			commentText = commentText +
+				fmt.Sprintf("\t* Had merge conflicts, you will have to fix this yourself :crying_cat_face:\n")
```
lluiscampos, line 253: Crying cat face, lol

cherrypicks.go
```diff
@@ -135,3 +141,136 @@ Hello :smile_cat: This PR contains changelog entries. Please, verify the need of
...
+		}
+	}
+	// Comment with cherry links on the PR
+	commentText := `Hi :smileycat:
+I did my very best, and this is the result of the cherry pick operation:
+`
+	for _, targetBranch := range targetBranches {
+		if !conflicts[targetBranch] {
+			commentText = commentText +
+				fmt.Sprintf("\t* %s:white_check_mark:\n", targetBranch)
```
lluiscampos, line 250: For consistency, just would add an space
```suggestion
				fmt.Sprintf("\t* %s :white_check_mark:\n", targetBranch)
```

cherrypicks.go
```diff
@@ -135,3 +141,136 @@ Hello :smile_cat: This PR contains changelog entries. Please, verify the need of
...
+		return fmt.Errorf("Failed to create the PR for: (%s) %v",
+			comment.GetRepo().GetName(), err)
+	}
+	return nil
+}
+
+func cherryPickPR(log *logrus.Entry, comment *github.IssueCommentEvent, pr *github.PullRequest, conf *config, body string, githubClient clientgithub.Client) error {
+	targetBranches := parseCherryTargetBranches(body)
+	if len(targetBranches) == 0 {
+		return fmt.Errorf("No target branches found in the comment body: %s", body)
```
lluiscampos, line 230: Thinking out loud, but this error, as well as "... is not a branch in..." from `cherryPickToBranch` would be nice to be relayed to the PR. We know that right now is quite hard to find the logs, so maybe we could find about a general way of the bot to always report comments.

In any case, let's just keep this in the back of our heads and not complicate (and add risk in) this PR.

cherrypicks_test.go
```diff
@@ -185,3 +185,32 @@ Hello :smile_cat: This PR contains changelog entries. Please, verify the need of
...
+	tests := map[string]struct {
+		input    string
+		expected []string
+	}{
+		"Success": {
+			input: `
+cherry pick to:
+    * 2.6.x
+    * 2.5.x
+    * 2.4.x
```
lluiscampos, line 199: Ah, now I understand the idea behind the regex!

Okey, but I was expecting something simpler: as I user I would have written "cherry pick to: 2.5.x and 2.4.x" and expect it to work, or "cherry pick to 2.6.x, 2.5.x". --> In other words, not being strict in having one version per line.

But I see that this makes the code simpler, so we can start as-is, I have no strong opinion.

main.go
```diff
@@ -130,7 +130,8 @@ const (
 )
 
 const (
-	commandStartPipeline = "start pipeline"
+	commandStartPipeline    = "start pipeline"
+	commandCherryPickBranch = "cherry pick to:"
```
lluiscampos, line 134: I personally always type "cherry-pick". Could it be possible to make the bot accept both space and `-`? Otherwise is just fine, I only need to get used to.

main_comment.go
```diff
@@ -89,6 +89,12 @@ func processGitHubComment(ctx *gin.Context, comment *github.IssueCommentEvent, g
 				log.Errorf("Could not start build: %s", err.Error())
 			}
 		}
+	case strings.Contains(body, commandCherryPickBranch):
+		log.Infof("Attempting to cherry-pick the changes in PR: %v", pr)
```
lluiscampos, line 93: Lol, I just wrote about the `-` in cherry-pick/cherry pick and I see that even bot itself is not consistent :P

lluiscampos (changes requested):
This looks really good! Wonderful work. :rainbow: 

The interface is good as is :heavy_check_mark: - To make it even more awesome, the bot could take the Mender versions as input and translate automatically to independent components versions. I know this complicates the code, but once we are on it we might want to make it easy-peasy for ourselves in the future :)

I am requesting changes mainly for the `git ls-remote` comment below, I am NOT requesting a change in the versions input (only suggesting).

Other than that few comments and questions below.

cherrypicks.go
```diff
@@ -135,3 +141,136 @@ Hello :smile_cat: This PR contains changelog entries. Please, verify the need of
...
+		"ls-remote",
+		"--heads",
+		"git@github.com:mendersoftware/"+repository+".git",
+	).With(&git.State{Dir: tdir}).Run()
+	if err != nil {
+		log.Error(err)
+		return false
+	}
+	for _, branch := range strings.Split(branches, "\n") {
+		if branch == queryBranch {
```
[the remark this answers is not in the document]
oleorhagen: Good catch!

cherrypicks.go
oleorhagen: Yes, good idea!

Let's kick the can down the road for now, but how about the error logs get batched, and then added, with a drop-down menu, like the build versions are? Then the comment is small, and the logs can be easily perused if you so wish?

cherrypicks_test.go
oleorhagen: Yes, I see what you mean, and I see that others have done so too, but let's keep it simple for now :)

If we get all the other parts to work, this should not be too hard to rework later on

main.go
oleorhagen: Agreed, `cherry-pick` it is ;)

oleorhagen:
> The interface is good as is heavy_check_mark - To make it even more awesome, the bot could take the Mender versions as input and translate automatically to independent components versions. I know this complicates the code, but once we are on it we might want to make it easy-peasy for ourselves in the future :)
> 
Great idea!

Implementing

cherrypicks.go
lluiscampos: That would look good, yes. As you said, let's think about it later down the road.

oleorhagen:
> > The interface is good as is heavy_check_mark - To make it even more awesome, the bot could take the Mender versions as input and translate automatically to independent components versions. I know this complicates the code, but once we are on it we might want to make it easy-peasy for ourselves in the future :)
> 
> Great idea!
> 
> Implementing

Added!

Also added a simple unit test, which helped catch quite a few issues actually!

I do not think that this will work OOTB still though, but it'll be significantly closer, so @lluiscampos please have a lookover when you have time :eye: :hourglass:  :smile_cat: 

I'm not rebasing until you've had a look, to keep the diff familiar :100:

lluiscampos (approved):
Looking good! Some comments, but this is a :green_circle: light from me.

Even if it is a bit daunting to just merge it, I actually think it is low risk. My reasoning is that this feature is comment-triggered and not webhook-triggered, so it shouldn't affect the other parts of the application.

cherrypicks.go
```diff
@@ -135,3 +141,178 @@ Hello :smile_cat: This PR contains changelog entries. Please, verify the need of
...
+}
+
+func cherryPickToBranch(log *logrus.Entry, comment *github.IssueCommentEvent, pr *github.PullRequest, conf *config, targetBranch string) (*github.PullRequest, error) {
+
+	if !branchIsABranchInRepository(log, comment.GetRepo().GetName(), targetBranch) {
+		return nil, fmt.Errorf("%s is not a branch in %s", targetBranch, comment.GetRepo().GetName())
+	}
+
+	prBranchName := fmt.Sprintf("cherry-%s-%s",
+		targetBranch, pr.GetHead().GetLabel())
```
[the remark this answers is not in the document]
oleorhagen: Nice catch!

PR #94, opened 2021-05-11
[gitlab_sync] Reuse state in subsequent git.Command(s)

lluiscampos:


lluiscampos:
I guess the nicest would have been to extend `Commands(cmds ...*gitCmd)` for accepting an existing state. But let's play it safe at this point :)

oleorhagen (approved):
Nice!

oleorhagen:
> I guess the nicest would have been to extend `Commands(cmds ...*gitCmd)` for accepting an existing state. But let's play it safe at this point :)

haha, exactly what went through my mind!

PR #97, opened 2021-06-09
Add monitor-client to watch repos list

lluiscampos:
And remove mender-python-client-dev

### DIFF
--- a/main.go
+++ b/main.go
@@ -105,8 +105,6 @@ var defaultWatchRepositoriesSync = []string{
 	// websites
 	"mender.io",
 	"northern-tech-web",
-	// tmp repo - it is private but will be merged/removed later
-	"mender-python-client-dev",
 }
 
 var qemuBuildRepositories = []string{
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -92,6 +92,7 @@ var defaultWatchRepositoriesSync = []string{
 	"devicemonitor",
 	// client
 	"mender-binary-delta",
+	"monitor-client",
```
merlin-northern, line 95: I hereby request a comment preceding this line:
```c
// Monitoring
```

merlin-northern (approved):
lgtm, Ole is much better expert here, so I also rely on his review.

main.go
```diff
@@ -105,8 +105,6 @@ var defaultWatchRepositoriesSync = []string{
 	// websites
 	"mender.io",
 	"northern-tech-web",
-	// tmp repo - it is private but will be merged/removed later
-	"mender-python-client-dev",
```
oleorhagen, line 109: Thx, I saw this the other day, and added it to my TODO list :)

oleorhagen (approved):

main.go
lluiscampos: ok

PR #105, opened 2021-06-28
Do not start pipeline when PR is coming from branch

lluiscampos:
* Fix pipeline ref when PR is coming from branch
head points to the PR source, base to the target to be merged to.

* Do not start pipeline when PR is coming from branch


lluiscampos:
The first commit fixes the issue that @mzedel and me have seen: every change on PRs coming from branches will trigger a pipeline with the base ref (aka master).

The second commit removes the functionality all together, as I realized that for branches the pipelines are run automatically when when sync using push event.

main_pullrequest.go
```diff
@@ -24,7 +24,7 @@ func processGitHubPullRequest(ctx *gin.Context, pr *github.PullRequestEvent, git
 	req := pr.GetPullRequest()
 	base := req.GetBase()
 	head := req.GetHead()
-	prRef = base.GetRef()
+	prRef = head.GetRef()
```
lluiscampos, line 28: @alfrunes Actually, after my second commit, `prRef` does not need to be defined here anymore. Should I clean it up?

alfrunes (approved):

PR #106, opened 2021-07-15
FIX: Remotes appearing in branchnames

alfrunes:
Though it's probably not the safest way to strip the "ref directory", the same logic is used elsewhere...


tranchitella (approved):

alfrunes:
FYI @lluiscampos

PR #120, opened 2021-10-05
Extract GitHub Organization from webhooks payload

lluiscampos:
Instead of hard-coding it to `mendersoftware`, so that it can be used by
other GitHub Organizations (namely: CFEngine).

### DIFF
--- a/cherrypicks_test.go
+++ b/cherrypicks_test.go
@@ -16,6 +16,9 @@ import (
 )
 
 func TestSuggestCherryPicks(t *testing.T) {
+
+	gitHubOrg := "mendersoftware"
+
 	testCases := map[string]struct {
 		pr      *github.PullRequestEvent
 		err     error
@@ -164,7 +167,7 @@ Hello :smile_cat: This PR contains changelog entries. Please, verify the need of
 					mock.MatchedBy(func(ctx context.Context) bool {
 						return true
 					}),
-					githubOrganization,
+					gitHubOrg,
 					*tc.pr.Repo.Name,
 					*tc.pr.Number,
 					tc.comment,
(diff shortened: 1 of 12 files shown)


olehermanse (approved):

util.go
```diff
@@ -34,3 +36,19 @@ func getRemoteURLGitLab(org, repo string) (string, error) {
 	}
 	return remoteURL, nil
 }
+
+func getGitHubOrganization(webhookType string, webhookEvent interface{}) (string, error) {
+	switch webhookType {
+	case "pull_request":
+		pr := webhookEvent.(*github.PullRequestEvent)
+		return *pr.Organization.Login, nil
```
tranchitella, line 44: Doesn't really matter too much, but doesn't the API allow you to use the `pr.GetOrganization.GetLogin()` syntax?

cherrypicks_test.go
```diff
@@ -164,15 +164,16 @@ Hello :smile_cat: This PR contains changelog entries. Please, verify the need of
 					mock.MatchedBy(func(ctx context.Context) bool {
 						return true
 					}),
-					githubOrganization,
+					"mendersoftware",
```
tranchitella, line 167: Maybe use a constant?

tranchitella (approved):

util.go
```diff
@@ -34,3 +36,19 @@ func getRemoteURLGitLab(org, repo string) (string, error) {
...
+	case "push":
+		push := webhookEvent.(*github.PushEvent)
+		return *push.Repo.Organization, nil
+	case "issue_comment":
+		comment := webhookEvent.(*github.IssueCommentEvent)
+		return *comment.Repo.Owner.Login, nil
```
lluiscampos, line 50: FYI, we should be able to get `Organization.Login` from all kind of events. I opened a PR upstream to have these objects available. See https://github.com/google/go-github/pull/2116

util.go
lluiscampos: Yes, thanks

tranchitella (approved):

PR #124, opened 2021-10-13
Add a comment for notifying up front which branches can be automatically targeted by the cherry-pick bot

oleorhagen:
Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -19,6 +19,15 @@ include:
   - project: 'Northern.tech/Mender/mendertesting'
     file: '.gitlab-ci-check-docker-deploy.yml'
 
+
+test:unit:
+ script:
+   - git config --global user.email test@example.com
+   - git config --global user.name  mendertester
+   - go list ./... | grep -v vendor | xargs -n1 -I {} go test -v -covermode=atomic -coverprofile=../../../{}/coverage.txt {} 2>&1 | tee /dev/stderr | go-junit-report > ${CI_PROJECT_DIR}/test-results.xml || exit $?
+   - mkdir -p tests/unit-coverage && find . -name 'coverage.txt' -exec cp --parents {} ./tests/unit-coverage \;
+   - tar -cvf ${CI_PROJECT_DIR}/unit-coverage.tar tests/unit-coverage
+
 test:acceptance_tests:
   stage: test
   tags:


.gitlab-ci.yml
```diff
@@ -19,6 +19,15 @@ include:
   - project: 'Northern.tech/Mender/mendertesting'
     file: '.gitlab-ci-check-docker-deploy.yml'
 
+
+test:unit:
+ script:
+   - git config --global user.email test@example.com
+   - git config --global user.name  mendertester
```
lluiscampos, line 26: Thinking out loud: wouldn't be cool that the template supports somewhat a "user hook" to run extra shell commands? We have something like that for OS dependencies, where user pipelines can have a `deb-requirements.txt` file that the pipeline installs if existing. So something similar for a user pipeline like this one to run the `git config` commands without needing to copy/paste the rest of the `script` (and risking becoming out of sync).

Anyway, just an idea for the future.

cherrypicks.go
```diff
@@ -100,7 +100,11 @@ func suggestCherryPicks(log *logrus.Entry, pr *github.PullRequestEvent, githubCl
 		if err != nil {
 			return err
 		} else if releaseBranch != "" {
-			releaseBranches = append(releaseBranches, releaseBranch+" (release "+version+")")
+			if isBottable(pr.GetRepo().GetName(), conf, pr.GetPullRequest(), releaseBranch) {
```
lluiscampos, line 103: I personally like the name, `isBottable`, but it really is not a very descriptive name, it should at least indicate that this is about cherry-picking and not any other "bot" actionable thing. Suggestion: `isCleanCherryPick`, or `isCherryPickBottable`.

cherrypicks.go
```diff
@@ -140,33 +144,48 @@ Hello :smile_cat: This PR contains changelog entries. Please, verify the need of
...
+func isBottable(repoName string, conf *config, pr *github.PullRequest, targetBranch string) bool {
+	_, state, err := cherryIsPortableNoConflicts(repoName, conf, pr, targetBranch)
+	state.Cleanup()
+	if err != nil {
+		logrus.Errorf("isBottable received error: %s", err.Error())
+	}
+	return err == nil
+}
 
+func cherryIsPortableNoConflicts(repoName string, conf *config, pr *github.PullRequest, targetBranch string) (string, *git.State, error) {
```
lluiscampos, line 156: (related to my previous comment) I would then name this one `tryCherryPickToBranch` or something like that, because we are both using it as a dry-run (from `isBottable` (sic)) and as an actual action from `cherryPickToBranch`.

cherrypicks_test.go
```diff
@@ -142,6 +142,34 @@ Hello :smile_cat: This PR contains changelog entries. Please, verify the need of
...
+			pr: &github.PullRequestEvent{
+				Action: github.String("closed"),
+				Number: github.Int(29),
+				PullRequest: &github.PullRequest{
+					Base: &github.PullRequestBranch{
+						Ref: github.String("master"),
+						SHA: github.String("4c6d93ba936031ee00d9c115ef2dc61597bc1296"),
+					},
+					Head: &github.PullRequestBranch{
+						Ref: github.String("logbuffering"),
```
lluiscampos, line 158: Will removing this branch in the future make the test break?

cherrypicks.go
oleorhagen: `cherryPickToBranch` it is

PR #136, opened 2021-11-08
Fix the release version sort

oleorhagen:
Previously we only  relied on the creator-date. This sorts the releases
semantically on the key values of the major and minor version of the release.

Changelog: None


oleorhagen:
Oh my, still won't work :(

PR #138, opened 2021-11-09
Freeze the integration repository used in tests

oleorhagen:
In order to always have a stable pipe, freeze the integration repository used,
through referring to the bot's own and unused fork of the repository.

Changelog: None

### DIFF
--- a/cherrypicks_test.go
+++ b/cherrypicks_test.go
@@ -181,7 +181,7 @@ Hello :smile_cat: This PR contains changelog entries. Please, verify the need of
 	}
 	defer os.RemoveAll(tmpdir)
 
-	gitSetup := exec.Command("git", "clone", "https://github.com/mendersoftware/integration.git", tmpdir)
+	gitSetup := exec.Command("git", "clone", "https://github.com/mender-test-bot/integration.git", tmpdir)
 	gitSetup.Dir = tmpdir
 	_, err = gitSetup.CombinedOutput()
 	if err != nil {


oleorhagen:
If you agree, I will leave a note in the integration repository (mender-test-bot), so that it remains untouched.

cherrypicks_test.go
```diff
@@ -181,7 +181,7 @@ Hello :smile_cat: This PR contains changelog entries. Please, verify the need of
 	}
 	defer os.RemoveAll(tmpdir)
 
-	gitSetup := exec.Command("git", "clone", "https://github.com/mendersoftware/integration.git", tmpdir)
+	gitSetup := exec.Command("git", "clone", "https://github.com/mender-test-bot/integration.git", tmpdir)
```
lluiscampos, line 184: I agree with the principle, but not in the actual implementation.

I think we should still clone `mendersoftware` repo, but then just checkout a known git sha. We can for example take today's master, and we will be updating it as we need (when for whatever reason we have a regression that needs a new test or a new feature that requires newer master)

lluiscampos (changes requested):
Good idea! See my comment below

cherrypicks_test.go
oleorhagen: But the new release will always mess up the result of running the tests. i.e., they will have to be upgraded with every release

lluiscampos: Discussed f2f: Okey! :+1:

lluiscampos (approved):

oleorhagen:
Resolved to using golden files instead.

PR #139, opened 2021-11-10
menderqa_pipeline: Base the watch repositories list in release_tool list

lluiscampos:
So that we remove one point of duplication in our ecosystem.

The list was only being used to trigger in which repos to offer the
"start pipeline" functionality. Instead of a full list, hard-code only
the repos that are not part of the tool (namely: meta-mender).

The only drawback is that we are now suggesting in `gui` to start the
pipeline, but this might come in handy in the future if we integrate the
e2e tests with mender-qa.

### DIFF
--- a/menderqa_pipeline.go
+++ b/menderqa_pipeline.go
@@ -45,7 +45,17 @@ func getBuilds(log *logrus.Entry, conf *config, pr *github.PullRequestEvent) []b
 
 	makeQEMU := false
 
-	for _, watchRepo := range conf.watchRepositoriesTriggerPipeline {
+	// we need to have the latest integration/master branch in order to use the release_tool.py
+	if err := updateIntegrationRepo(conf); err != nil {
+		log.Warnf(err.Error())
+	}
+
+	watchRepositoriesTriggerPipeline, err := getListOfWatchedRepositories(conf)
+	if err != nil {
+		log.Warnf(err.Error())
+	}
+
+	for _, watchRepo := range watchRepositoriesTriggerPipeline {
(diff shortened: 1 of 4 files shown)


oleorhagen:
@lluiscampos if you rebase on master now, this should be good :)

oleorhagen:
whops

oleorhagen:
Well, looks legit actually, but should be easy to fix :)

menderqa_pipeline.go
```diff
@@ -58,11 +68,6 @@ func getBuilds(log *logrus.Entry, conf *config, pr *github.PullRequestEvent) []b
 				}
 			}
 
-			// we need to have the latest integration/master branch in order to use the release_tool.py
-			if err := updateIntegrationRepo(conf); err != nil {
```
oleorhagen, line 62: Question:
Why was this needed? :thinking:

oleorhagen (approved):
Great improvement!

menderqa_pipeline.go
lluiscampos: It would work _most of the times_ keeping it inside the loop.

But the detail is that the day that you add a repo, you need to pull integration before being able to know if it is a repo that needs watching or not. Previously the list was hardcoded so this was not needed.

One might argue that this is slightly less efficient as we are pulling always when before we only pulled once we knew the repo was one to track. But well... we are not optimizing for runtime here, are we?

oleorhagen: Good to know, thanks, makes sense.
Nah, no worries about runtime here.

PR #141, opened 2021-11-12
Move the acceptance tests to follow a golden file test style

oleorhagen:
Changelog: None

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -1,3 +1,5 @@
 #!/bin/sh
 
+pip install pytest-golden
+
 py.test -s --tb=short -vv /testing/tests "$@"
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,10 @@ acceptance-testing-up:
 acceptance-testing-run:
 	docker exec tests_acceptance-testing_1 /testing/run.sh
 
+.PHONY: acceptance-testing-update-golden-files
+acceptance-testing-update-golden-files:
+	docker exec tests-acceptance-testing-1 /testing/run.sh --update-goldens
+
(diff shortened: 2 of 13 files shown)


oleorhagen:
Obviously this could be unified into a single test function now, but i decided against it in order to increase the visibility.

tests/run.sh
```diff
@@ -1,3 +1,5 @@
 #!/bin/sh
 
-py.test -s --tb=short -vv /testing/tests "$@"
+pip install pytest-golden
+
+ py.test -s --tb=short -vv /testing/tests "$@"
```
lluiscampos, line 5: nitpick: unnecessary space here

tests/run.sh
```diff
@@ -1,3 +1,5 @@
 #!/bin/sh
 
-py.test -s --tb=short -vv /testing/tests "$@"
+pip install pytest-golden
```
lluiscampos, line 3: This should be in the `Dockerfile` instead.

lluiscampos (approved):
Amazing! :raised_hands: 

I had to read-up a bit on `pytest-golden` but it seems to be very wise choice. And seeing the pipeline green for the first time in a while is very encouraging too :)

I haven't manually checked the golden files, obviously, so I just trust you and the tool on that.

tests/run.sh
oleorhagen:
```
    image: mendersoftware/mender-test-containers:acceptance-testing
```
So I would have to add it to this image, which I decided against, considering the use case. But if you insist.

lluiscampos: Okey, I don't insist :)

PR #144, opened 2021-11-25
QA-336 Create the pr_ branches with exceptions

merlin-northern:
ChangeLog:title
Signed-off-by: Peter Grzybowski <peter@northern.tech>

### DIFF
--- a/main_pullrequest.go
+++ b/main_pullrequest.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"regexp"
 	"strings"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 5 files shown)


merlin-northern:
Kristian: if you would be so kind and give me an early review here, is this the right direction? before I start adding tests.

merlin-northern:
> Kristian: if you would be so kind and give me an early review here, is this the right direction? before I start adding tests.

@kacf thanks for early feedback over slack. I hope it makes sense. do you think we need more tests around it?

main_pullrequest.go
```diff
@@ -11,8 +12,32 @@ import (
 	clientgithub "github.com/mendersoftware/integration-test-runner/client/github"
 )
 
-func processGitHubPullRequest(ctx *gin.Context, pr *github.PullRequestEvent, githubClient clientgithub.Client, conf *config) error {
+func startPipeline(branchName string) bool {
```
kacf, line 15: Suggest renaming to `shouldStartPipeline`. The original name makes it sound like you are starting the pipeline (and this is what I first thought when I read it out of context below).

kacf (approved):
Just a minor suggestion, it looks good to me!

PR #160, opened 2022-02-03
fix: cherry-picks to standard branches now supported

oleorhagen:
Previously cherry-picks were only supported in release branches matching some
release version, like `1.2.x`.

Now, this functionality is extended to the standard branchnames, like:

* master
* hosted
* staging
* production

Changelog: None

### DIFF
--- a/cherrypicks.go
+++ b/cherrypicks.go
@@ -332,7 +332,7 @@ I did my very best, and this is the result of the cherry pick operation:
 
 func parseCherryTargetBranches(body string) []string {
 	matches := []string{}
-	regex := regexp.MustCompile(` *\* *([0-9]+\.[0-9]+\.x)`)
+	regex := regexp.MustCompile(` *\* *(([[:word:]]+[_\.-]?)+)`)
 	for _, line := range strings.Split(body, "\n") {
 		if m := regex.FindStringSubmatch(line); len(m) > 1 {
 			matches = append(matches, m[1])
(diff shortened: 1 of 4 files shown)


oleorhagen:
I'm guessing, we really could open this up to any branch name?

What do you think?

lluiscampos:
> I'm guessing, we really could open this up to any branch name?
> 
> What do you think?

yes, why not?

lluiscampos (approved):
Looks good

But I think we could remove the check all together. We usually know what we are doing when asking for a cherry-pick, and we still will review every PR before merge. So I don't see any risk in allowing any arbitrary branch.

lluiscampos (approved):

lluiscampos:
@oleorhagen I took the liberty of fixing the unit tests and force-push to your branch. I want to merge this to unblock another PR.

Side note, the three test cases of `TestSuggestCherryPicks` are really the same now, should we unify in one?

oleorhagen:
> @oleorhagen I took the liberty of fixing the unit tests and force-push to your branch. I want to merge this to unblock another PR.
> 
> Side note, the three test cases of `TestSuggestCherryPicks` are really the same now, should we unify in one?

Alright, nice!

I was actually planning on waiting for your release, but same matter. Or rather, even better actually ;)

oleorhagen:
> Side note, the three test cases of TestSuggestCherryPicks are really the same now, should we unify in one?

Sure

PR #162, opened 2022-02-11
Adds new repo/pull/<pr_number> syntax to integration-test-runner

MaciejTe:
Adds new repo/pull/<pr_number> syntax to integration-test-runner

Changelog: None

Signed-off-by: Maciej Tomczuk <maciej.tomczuk@northern.tech>

### DIFF
--- a/main_comment.go
+++ b/main_comment.go
@@ -37,14 +37,14 @@ func processGitHubComment(
 	}
 
 	// but ignore comments from myself
-	if comment.Sender.GetLogin() == "mender-test-bot" {
+	if comment.Sender.GetLogin() == githubBotName {
 		log.Warnf("%s commented, probably giving instructions, ignoring", comment.Sender.GetLogin())
 		return nil
 	}
 
 	// filter comments mentioning the bot
-	body := comment.Comment.GetBody()
-	if !strings.Contains(body, "@"+githubBotName) {
+	commentBody := comment.Comment.GetBody()
+	if !strings.Contains(commentBody, "@"+githubBotName) {
 		log.Info("ignoring comment not mentioning me")
 		return nil
 	}
@@ -76,11 +76,11 @@ func processGitHubComment(
 
 	// extract the command and check it is valid
 	switch {
-	case strings.Contains(body, commandStartPipeline):
+	case strings.Contains(commentBody, commandStartPipeline):
 		// make sure we only parse one pr at a time, since we use release_tool
 		mutex.Lock()
 
-		prsRepos, err := parsePrOptions(body)
+		prsRepos, err := parsePrOptions(commentBody)
 		// get the list of builds
 		prRequest := &github.PullRequestEvent{
 			Repo:        comment.GetRepo(),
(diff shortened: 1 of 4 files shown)


main_comment_test.go
```diff
@@ -349,6 +349,22 @@ func TestParsePrOptions(t *testing.T) {
...
+			},
+		},
+		"start pipeline with --pr flags (syntax without 'pull' and 'head')": {
+			StartPipelineComment: "start pipeline --pr mender-connect/88 --pr deviceconnect/12 --pr mender/3.1.x",
+			RepoToPr: map[string]string{
+				"mender-connect": "pull/88/head",
+				"deviceconnect":  "pull/12/head",
+				"mender":         "3.1.x",
+			},
+		},
```
kacf, line 367: Please add a `--pr mender/feature-my-feature-branch` as well, so we have that tested once and for all.

kacf (changes requested):

main_comment.go
```diff
@@ -164,6 +164,9 @@ func parsePrOptions(commentBody string) (map[string]string, error) {
 				if _, err := strconv.Atoi(revision); err == nil {
 					revision = "pull/" + revision + "/head"
 				}
+				if !strings.Contains(revision, "/head") && !strings.Contains(revision, ".") {
+					revision = revision + "/head"
+				}
```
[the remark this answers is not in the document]
MaciejTe: I resigned from regexes and refactored the whole function.

main_comment_test.go
MaciejTe: Done!

main_comment.go
```diff
@@ -147,30 +147,40 @@ func processGitHubComment(
...
+			var revision string
+			switch len(userInputParts) {
+			case 2: // we can have both deviceauth/1 and mender/3.1.x syntax
+				// repo/<pr_number> syntax
+				if _, err := strconv.Atoi(userInputParts[1]); err == nil {
+					revision = "pull/" + userInputParts[1] + "/head"
+				} else {
+					// feature branch
+					prPartsAndRandString := strings.Split(userInputParts[1], " ")
+					revision = prPartsAndRandString[0]
```
kacf, line 171: I think iterating over the words as in the original version is better, mainly because the new version makes it hard to add new arguments. What if we add an `--enterprise` argument? It also leads to splitting on spaces in the above line, long after you already split on `--pr` further above, which is a bit unexpected. It seems more natural to split on word boundaries first, and then inspect each word separately.

Otherwise I agree with the semantics. This is the behavior we want.

MaciejTe: I came back to iterating over words.

main_comment.go
```diff
@@ -147,31 +147,40 @@ func processGitHubComment(
...
+			if len(userInput) > 0 {
+				var revision string
+				switch len(userInputParts) {
+				case 2: // we can have both deviceauth/1 and mender/3.1.x syntax
+					// repo/<pr_number> syntax
+					if _, err := strconv.Atoi(userInputParts[1]); err == nil {
+						revision = "pull/" + userInputParts[1] + "/head"
+					} else {
+						// feature branch
+						prPartsAndRandString := strings.Split(userInputParts[1], " ")
```
kacf, line 167: I don't understand why splitting on spaces here is necessary. There should not be any, or? Is it just a leftover from the previous implementation? Strictly speaking it does no harm, it's just unnecessary, and you should be able to just write this:

```golang
revision = userInputParts[1]
```

main_comment_test.go
```diff
@@ -349,6 +349,23 @@ func TestParsePrOptions(t *testing.T) {
 				"mender":         "3.1.x",
 			},
 		},
+		"start pipeline with --pr flags (new syntax)": {
+			StartPipelineComment: "start pipeline --pr mender-connect/pull/88 --pr deviceconnect/pull/12 --pr mender/3.1.x --pr deviceauth/feature-branch",
+			RepoToPr: map[string]string{
+				"mender-connect": "pull/88/head",
+				"deviceconnect":  "pull/12/head",
+				"mender":         "3.1.x",
+				"deviceauth":     "feature-branch",
```
kacf, line 358: 👍

kacf (approved):
Thanks for sticking with it! I have one comment, where I think you can simplify, but it's not overly important, so I'm approving, and you can decide what to do!

main_comment.go
MaciejTe: Good catch, thank you!

PR #163, opened 2022-02-14
test-bot to support single line syntax for cherry-picking

snusifer:
Currently, our test-bot (integration-test-runner) repository does only support the syntax:

@mender-test-bot cherry-pick to:
* master
* hosted

This PR enables support for one-liner syntax:
``@mender-test-bot cherry-pick to: `master` `hosted` ``

Signed-off-by: Mikael Torp-Holte [mikael.torp-holte@northern.tech](mailto:mikael.torp-holte@northern.tech)


cherrypicks.go
```diff
@@ -340,3 +349,14 @@ func parseCherryTargetBranches(body string) []string {
 	}
 	return matches
 }
+
+func parseCherryTargetBranchesSingleLine(body string) []string {
+	matches := []string{}
+	regex := regexp.MustCompile(`\x60(([[:word:]]+[_\.-]?)+)\x60`)
+	for _, m := range regex.FindAllStringSubmatch(body, -1) {
+		if len(m) > 1 {
+			matches = append(matches, m[1])
```
oleorhagen, line 358: Question:

Can we not simply append all the matches in one go? :smile_cat:

cherrypicks_test.go
```diff
@@ -398,7 +398,45 @@ func TestParseCherryTargetBranches(t *testing.T) {
...
+		"feature-branch": {
+			body:     "cherry-pick to: `feature-independe_testing-1`",
+			expected: "feature-independe_testing-1",
+		},
+		"1.2.x": {
+			body:     "cherry-pick to: `1.2.x`",
+			expected: "1.2.x",
+		},
+		"1.2.x with escape char": {
+			body:     "cherry-pick to: `1.2.x`" + `\r`,
```
oleorhagen, line 432: Suggestion: 
Add one test with multiple branches on the same line :)

oleorhagen (commented):
Looks nice I think!

Checkout https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md for what we do with signoffs.

Also the git history needs to be squashed into one commit, and I am gonna need a better explanation of the change in the commit message :)

But looking real good!

cherrypicks_test.go
snusifer: I agree! The original test didn't do it so I thought it might not be necessary. Maybe do this for both parse-tests?

cherrypicks.go
oleorhagen: Maybe `FindAllString()` will be better here actually? :thinking: 

It is really closer to what we want I think

cherrypicks_test.go
```diff
@@ -394,11 +394,60 @@ func TestParseCherryTargetBranches(t *testing.T) {
...
+		* hosted
+		* example-branch`,
+			expected: "master hosted example-branch",
+		},
+	}
+
+	for name, test := range tests {
+		t.Log(name)
+		res, _ := parseCherryTargetBranches(test.body)
+		assert.Equal(t, test.expected, strings.Join(res, " "))
```
oleorhagen, line 409: Suggestion:

Change the type to the type returned by the function instead of joining here. It really is what we want to test.

cherrypicks_test.go
```diff
@@ -394,11 +394,60 @@ func TestParseCherryTargetBranches(t *testing.T) {
...
+		"1.2.x": {
+			body:     "cherry-pick to: `1.2.x`",
+			expected: "1.2.x",
+		},
+		"1.2.x with escape char": {
+			body:     "cherry-pick to: `1.2.x`" + `\r`,
+			expected: "1.2.x",
+		},
+		"multiple branches": {
+			body:     "cherry-pick to: `master` `hosted` `example-branch`",
```
oleorhagen, line 443: :tada: 

Question:

Will this work if I add, say, a `,` in between the branches?

cherrypicks_test.go
```diff
@@ -247,7 +247,7 @@ cherry pick to:
 
 	for name, test := range tests {
 		t.Log(name)
-		output := parseCherryTargetBranches(test.input)
+		output, _ := parseCherryTargetBranches(test.input)
```
oleorhagen, line 250: :+1:

cherrypicks.go
snusifer: I agree, but we are filtering the branches based on length. If it has length < 2, then we simply drop it, as we do in the original syntax parser function. Don't think golang has any means of filtering arrays like this unless we write our own. Hence, I don't think we are able to append in one go :/

cherrypicks_test.go
snusifer: I guess so, I thought perhaps having `expected: []string {"expected-output"}` for each entry would be a pain to the eyes.

cherrypicks_test.go
snusifer: Yes, it will also work if you don't. 

There are also other quirks as well. For instance, this will work:
`` @mender-test-bot `branch1` cherry-pick to: `branch2` ``
Easily fixable by having a regex look-behind, but I thought perhaps no one would ever write something like this. 

Please tell me if you would like the syntax to be "stricter" in those regards, or any other for that matter :)

cherrypicks.go
snusifer: Actually, I'm not too sure. `FindAllString` doesn't care about capturing groups (we don't want to include the backticks), and I don't think any functions does apart from those that contains the word `Submatch`. One must loop through the 2d array to retrieve the correct submatch I believe (go also doesnt support lookbehinds and lookaheads for complexity reasons)

oleorhagen (commented):
Looking good!

Some git-history fixup remaining :)

cherrypicks.go
oleorhagen: Probably explains why it was not used  in the first instance :)

cherrypicks_test.go
snusifer: ``@mender-test-bot `branch1` cherry-pick to: `branch2` ``will not work anymore now tho

oleorhagen:
Looks :green_book: to me :smile:

oleorhagen:
Good work @snusifer !

oleorhagen (approved):

PR #166, opened 2022-03-21
tests/golden-files: Update introducing mender-gateway

lluiscampos:


Makefile
```diff
@@ -54,7 +54,7 @@ acceptance-testing-run:
 
 .PHONY: acceptance-testing-update-golden-files
 acceptance-testing-update-golden-files:
-	docker exec tests-acceptance-testing-1 /testing/run.sh --update-goldens
+	docker exec tests_acceptance-testing_1 /testing/run.sh --update-goldens
```
oleorhagen, line 57: This is actually not a typo, but a difference depending on which version of docker you are running

oleorhagen (approved):

Makefile
lluiscampos: Okey, I'll drop this commit. I guess we should update the Makefile for all mentions to be one or the other syntax.

oleorhagen: Yeh, really it should detect the name through other means, like `ps`, and then use the name as a variable, or subshell output

PR #185, opened 2022-09-21
Test-runner crush fix and added PR comment updates on failures

0lmi:
chore: update PR with mindful message on failures
Ticket: QA-346

fix: test-runner crashing
Ticket: MEN-5881

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,7 +7,6 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
-	"reflect"
 	"sync"
 	"time"
 
@@ -138,10 +137,19 @@ func getConfig() (*config, error) {
 
 func getCustomLoggerFromContext(ctx *gin.Context) *logrus.Entry {
 	deliveryID, ok := ctx.Get("delivery")
-	if !ok || reflect.TypeOf(deliveryID).Kind() != reflect.String {
-		return nil
+	if !ok || !isStringType(deliveryID) {
+		return logrus.WithField("delivery", "nil")
 	}
 	return logrus.WithField("delivery", deliveryID)
+}
+
+func isStringType(i interface{}) bool {
+	switch i.(type) {
+	case string:
+		return true
+	default:
+		return false
+	}
 }
 
 func processGitHubWebhookRequest(
--- a/main.go
+++ b/main.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"reflect"
 	"sync"
 	"time"
 
@@ -137,7 +138,7 @@ func getConfig() (*config, error) {
 
 func getCustomLoggerFromContext(ctx *gin.Context) *logrus.Entry {
 	deliveryID, ok := ctx.Get("delivery")
-	if !ok {
+	if !ok || reflect.TypeOf(deliveryID).Kind() != reflect.String {
 		return nil
 	}
 	return logrus.WithField("delivery", deliveryID)
--- a/main_pullrequest.go
+++ b/main_pullrequest.go
@@ -42,9 +42,14 @@ func processGitHubPullRequest(
 	log.Debugf("Processing pull request action %s", action)
 	switch action {
 	case "opened", "edited", "reopened", "synchronize", "ready_for_review":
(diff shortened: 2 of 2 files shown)


main_pullrequest.go
```diff
@@ -42,9 +42,14 @@ func processGitHubPullRequest(
 	log.Debugf("Processing pull request action %s", action)
 	switch action {
 	case "opened", "edited", "reopened", "synchronize", "ready_for_review":
+		msgDetails := "see <a href=\"https://console.cloud.google.com/kubernetes/" +
+			"deployment/us-east1/company-websites/default/test-runner-mender-io/logs?" +
+			"project=gp-kubernetes-269000\">logs<\a> for details."
```
lluiscampos, line 47: This does not render as you expect
image

Use markdown instead

main_pullrequest.go
```diff
@@ -42,9 +42,14 @@ func processGitHubPullRequest(
...
 		// We always create a pr_* branch
 		if prRef, err = syncPullRequestBranch(log, pr, conf); err != nil {
 			log.Errorf("Could not create PR branch: %s", err.Error())
+			msg := "Kitty, there was an error syncing branches, " + msgDetails
```
lluiscampos, line 51: Kitty? Did you mean `:smile_cat:`?

lluiscampos: In any case I don't think we need it. Instead I would add a danger or problem kind of emoji.

lluiscampos (changes requested):

main.go
```diff
@@ -137,8 +137,8 @@ func getConfig() (*config, error) {
 
 func getCustomLoggerFromContext(ctx *gin.Context) *logrus.Entry {
 	deliveryID, ok := ctx.Get("delivery")
-	if !ok {
-		return nil
+	if !ok || deliveryID == nil {
+		return logrus.WithField("delivery", "nil")
```
lluiscampos, line 141: I don't think that this will fix the issue. As you said judging  by the docs it won't be `nil` unless `ok` if false. My theory is that `deliveryID` is _something_ but not an string. Then passing it down the road at some point breaks.

Maybe we can check if it's an string, and when not do this `return logrus.WithField("delivery", "nil")` - which also will work in the nil case.

main_pullrequest.go
0lmi: This should work with html. Here it's the typo - `\a` instead of `/a`.

main_pullrequest.go
0lmi: I was just following the ticket's description, IMO both works fine, but if you think removing it is better, then let's change that! 🙃

main.go
```diff
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"reflect"
```
lluiscampos, line 10: We don't need `reflect` package for this, we can achieve the same with a core functionality [type switch](https://go.dev/tour/methods/16).

lluiscampos (changes requested):

main_pullrequest.go
lluiscampos: Haha, I didn't expect anyone writing such a thing in the acceptance criteria... :sweat_smile:

main.go
0lmi: Ok. I thought with `reflect` it looks better, but you know better 🙃

main.go
```diff
@@ -137,12 +137,21 @@ func getConfig() (*config, error) {
 
 func getCustomLoggerFromContext(ctx *gin.Context) *logrus.Entry {
 	deliveryID, ok := ctx.Get("delivery")
-	if !ok {
+	if !ok || !isStringType(deliveryID) {
 		return nil
```
lluiscampos, line 141: But wait, this cannot return `nil` - wasn't that the root cause of the crash? You should keep your first approach of `return logrus.WithField("delivery", "nil")`

I am sorry I didn't catch this in the previous round when it was removed.

lluiscampos (approved):
Pre-approving, but please see my comment below.

main.go
0lmi: I understood your comment and it can and should be nil 🙃

PR #187, opened 2022-09-26
chore: Fix unit tests after Mender 3.4.0 release

lluiscampos:


snusifer (approved):
Perhaps we could think about ways to not do this every release? Cherry-pick branches are now deduced from `releases/versions.json` so if anything ought to break, it should be tests against that URL

cherrypicks_test.go
```diff
@@ -121,9 +121,9 @@ Hello :smile_cat: This PR contains changelog entries. Please, verify the need of
 			comment: &github.IssueComment{
 				Body: github.String(`
 Hello :smile_cat: This PR contains changelog entries. Please, verify the need of backporting it to the following release branches:
+2.1.x (release 3.4.x)
 2.0.x (release 3.3.x)
 2.0.x (release 3.2.x)
```
oleorhagen, line 126: I guess, technically `3.2` is no longer supported

oleorhagen (approved):

oleorhagen:
Thnks for fixing

cherrypicks_test.go
lluiscampos: I checked this with @snusifer and the issue is that the unit tests do not fetch `releases/versions.js` and instead fallback to the old method of using `release_tool`. That is why there are present here.

lluiscampos:
> Perhaps we could think about ways to not do this every release? Cherry-pick branches are now deduced from `releases/versions.json` so if anything ought to break, it should be tests against that URL (all though we don't have any as of now)

Yes, this will be an nice addition. Regardless of modifying it after every release, the point is that now we are testing a different flow that the one actually used.

PR #193, opened 2022-12-07
Add special handling for feature-c++-client branch.

kacf:
Ticket: MEN-5797

Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>


kacf:
The coveralls report says "We're sorry, this functionality is temporarily unavailable", so it's something on their side.

menderqa_pipeline.go
```diff
@@ -441,30 +441,58 @@ func getBuildParameters(
 
 	// set the rest of the CI build parameters
 	runIntegrationTests := "true"
+	runBackendIntegrationTests := "true"
```
oleorhagen, line 444: Are we adding support for this thing now?

oleorhagen (approved):

menderqa_pipeline.go
kacf: Not generic support, no. It's just to omit these builds from the C++ client builds.

PR #195, opened 2022-12-20
chore: Add the BUILD_CPP_CLIENT environment variable

oleorhagen:
Just pass some condition to mender-qa, so that we can conditionally build the cpp client recipe in Yocto.

### DIFF
--- a/menderqa_pipeline.go
+++ b/menderqa_pipeline.go
@@ -491,6 +491,10 @@ func getClientBuildParameters(
 		// that we can at least build it. MEN-6116 exists to clean this
 		// up later.
 		qemuParam = "false"
+		buildParameters = append(
+			buildParameters,
+			&gitlab.PipelineVariable{Key: "BUILD_CPP_CLIENT", Value: "true"},
+		)
 	}
 
 	buildParameters = append(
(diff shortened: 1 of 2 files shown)


menderqa_pipeline.go
```diff
@@ -491,6 +491,10 @@ func getClientBuildParameters(
 		// that we can at least build it. MEN-6116 exists to clean this
 		// up later.
 		qemuParam = "false"
+		buildParameters = append(
+			buildParameters,
+			&gitlab.PipelineVariable{Key: "BUILD_CPP_CLIENT", Value: "true"},
```
kacf, line 496: Is it not possible to determine this from the branch or version? I suppose that with a `meta-mender/feature-c++-client` branch, we can default to always build the C++ version, and then we shouldn't need this, or?

oleorhagen: I thought the plan was to merge the `meta-mender` branch to `master`, and not rely on a feature branch?

kacf: Hmm. Honestly I don't know which is better, I can see pros and cons with both approaches. Maybe having a branch is a bit safer though? To not stir up too much trouble while we're still working on this?

oleorhagen: Ref: f2f discussion

We will go down this route for now.

PR #201, opened 2023-01-03
chore: Use complete version for the Docker base images

lluiscampos:
For some reason dependabot concludes that "Latest version is 1.14-alpine3.11" when parsing the `Dockerfile`, which I suspect is due to the bugfix version not being present.


lluiscampos:
I want to merge this one without actually updating the versions, so that we can verify that dependabot parses it correctly afterwards.

mzedel (approved):
that is indeed strange, IMO that should have been picked up already, since there are many newer versions available...

let's see if this is maybe a bug in dependabot...

PR #209, opened 2023-01-12
INF-3042: ci: don't update repos-sync.cfengine.com to latest version

0lmi:
It's a workaround to address CFEngine team complaines about latest changes in integration-test-runner functionality, until required functionality is in place.

Changelog: None
Ticket: INF-3042
Signed-off-by: Alex Miliukov <oleksandr.miliukov@northern.tech>

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -93,4 +93,4 @@ publish:acceptance:
 
 sync:image:
   variables:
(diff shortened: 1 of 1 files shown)


olehermanse:
I have made a different workaround, changing the webhook, and it seems to work fine. See: https://tracker.mender.io/browse/SEC-763

olehermanse (changes requested):
Since we made another workaround, I don't think it's worth the risk of doing this change now. Let's wait and see / wait for the proper fix.

PR #211, opened 2023-01-24
QA-518: add list of repos to sync option

0lmi:
Ticket: [QA-518](https://tracker.mender.io/browse/QA-518)

With the change the following is achieved:
* it's possible to disable different features if some is not needed
* it's possible to set a list of managed repos and ignore the rest

The change is backward compatible and won't brake current deployments, but will allow to customise a deployment for CFEngine GitHub organisation.

Straight goal is to use the integration-test-runner to only sync specified list of repos for CFEngine GitHub organisation. With the implementation it can be achieved with using the following setup: `SYNC_REPOS_LIST="website,build"`, `NOT_PROCESS_PR_EVENTS="true"` and `NOT_PROCESS_COMMENTS="true"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,21 +2,20 @@
 
 ## Main features
 
-### `mender-qa` Pipeline run
-
-Set `WATCH_REPOS_PIPELINE` for the list of repositories for which to run
-`mender-qa` Pipeline. See source code for defaults.
-
-This is a Mender specific feature.
-
 ### GitHub -> GitLab sync
 
-All repositories from the configured GitHub organization are synced with GitLab.
+By default all repositories from the configured GitHub organization are synced with GitLab. To select a subset of repositories to sync, set `SYNC_REPOS_LIST` env variable with a comma separated list of repositories.
 
 ### GitLab PR branches
 
-For all repositories in the organization, a pr_XXX branch will be created in
-GitLab for every pull/XXX PR from GitHub.
+For all repositories in the organization, a pr_XXX branch will be created in GitLab for every pull/XXX PR from GitHub.
+
+### Processing GitHub events
+
+Currently the following GitHub events are processed:
+* `pull_request`: enabled by default, `DISABLE_PR_EVENTS_PROCESSING` disables the processing
+* `push`: enabled by default, `DISABLE_PUSH_EVENTS_PROCESSING` disables the processing
+* `issue_comment`: enabled by default, `DISABLE_COMMENT_EVENTS_PROCESSING` disables the processing
 
 ## Infrastructure
 
--- a/main.go
+++ b/main.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"strings"
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
@@ -32,6 +33,7 @@ type config struct {
 	isProcessPushEvents    bool
 	isProcessPREvents      bool
 	isProcessCommentEvents bool
+	reposSyncList          []string
 }
 
 type buildOptions struct {
@@ -88,13 +90,13 @@ const (
 )
 
 func getConfig() (*config, error) {
+	var reposSyncList []string
 	dryRunMode := os.Getenv("DRY_RUN") != ""
 	githubSecret := os.Getenv("GITHUB_SECRET")
 	githubToken := os.Getenv("GITHUB_TOKEN")
 	gitlabToken := os.Getenv("GITLAB_TOKEN")
 	gitlabBaseURL := os.Getenv("GITLAB_BASE_URL")
 	integrationDirectory := os.Getenv("INTEGRATION_DIRECTORY")
-	logLevel, found := os.LookupEnv("INTEGRATION_TEST_RUNNER_LOG_LEVEL")
 
 	//
 	// Currently we don't have a distinguishment between GitHub events and features.
@@ -110,8 +112,8 @@ func getConfig() (*config, error) {
 	// default: process comment events if not explicitly disabled
 	isProcessCommentEvents := os.Getenv("DISABLE_COMMENT_EVENTS_PROCESSING") == ""
 
+	logLevel, found := os.LookupEnv("INTEGRATION_TEST_RUNNER_LOG_LEVEL")
 	logrus.SetLevel(logrus.InfoLevel)
-
 	if found {
 		lvl, err := logrus.ParseLevel(logLevel)
 		if err != nil {
@@ -123,6 +125,12 @@ func getConfig() (*config, error) {
 			logrus.Infof("Set 'LogLevel' to %s", lvl)
 			logrus.SetLevel(lvl)
 		}
+	}
+
+	// Comma separated list of repos to sync (GitHub->GitLab)
+	reposSyncListRaw, found := os.LookupEnv("SYNC_REPOS_LIST")
+	if found {
+		reposSyncList = strings.Split(reposSyncListRaw, ",")
 	}
 
 	switch {
@@ -149,6 +157,7 @@ func getConfig() (*config, error) {
 		isProcessPushEvents:    isProcessPushEvents,
 		isProcessPREvents:      isProcessPREvents,
 		isProcessCommentEvents: isProcessCommentEvents,
+		reposSyncList:          reposSyncList,
 	}, nil
 }
 
--- a/main.go
+++ b/main.go
@@ -21,14 +21,17 @@ import (
 )
 
 type config struct {
-	dryRunMode           bool
-	githubSecret         []byte
-	githubProtocol       gitProtocol
-	githubOrganization   string
-	githubToken          string
-	gitlabToken          string
-	gitlabBaseURL        string
-	integrationDirectory string
+	dryRunMode             bool
+	githubSecret           []byte
+	githubProtocol         gitProtocol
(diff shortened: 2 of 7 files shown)


olehermanse:
> Straight goal is to use the integration-test-runner to only sync specified list of repos for CFEngine GitHub organisation. With the implementation it can be achieved with using the following setup: `SYNC_REPOS_LIST="website,build"`, `NOT_PROCESS_PR_EVENTS="true"` and `NOT_PROCESS_COMMENTS="true"`

@0lmi FYI, those 2 repo names are not correct, correct names are website and build-website.

olehermanse (commented):
Thank you for working on this! Looks good so far, but I cannot say I completely understand everything going on here.

main.go
```diff
@@ -85,16 +90,22 @@ const (
...
+	var reposSyncList []string
 	dryRunMode := os.Getenv("DRY_RUN") != ""
 	githubSecret := os.Getenv("GITHUB_SECRET")
 	githubToken := os.Getenv("GITHUB_TOKEN")
 	gitlabToken := os.Getenv("GITLAB_TOKEN")
 	gitlabBaseURL := os.Getenv("GITLAB_BASE_URL")
 	integrationDirectory := os.Getenv("INTEGRATION_DIRECTORY")
-	logLevel, found := os.LookupEnv("INTEGRATION_TEST_RUNNER_LOG_LEVEL")
+	// default: sync repos if not explicitly disabled
+	isSyncRepos := os.Getenv("NOT_SYNC_REPOS") == ""
```
lluiscampos, line 101: These variables need to be documented in the README.

lluiscampos (changes requested):
I think the direction of this PR is a bit off.

What CFEngine needs is:
* sync certain repos on push events (which we have covered with the current implementation)
* but also start pipelines on PR for that repos (which is not happening with the current implementation).

This relates to my comment below, that filtering by type of the webhook is just not a good abstraction, we need to filter by logical features, not from where these are triggered.

We can have a f2f conversation on this tomorrow :telephone_receiver:

main.go
```diff
@@ -85,16 +90,22 @@ const (
...
+	// default: process PR events if not explicitly disabled
+	isProcessPREvents := os.Getenv("NOT_PROCESS_PR_EVENTS") == ""
+	// default: process PR comments if not explicitly disabled
+	isProcessComments := os.Getenv("NOT_PROCESS_COMMENTS") == ""
```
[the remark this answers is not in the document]
0lmi: The high level idea of the implementation came from a discussion with @olehermanse.

> when setting up a hook in GitHub you can configure it to send only a subset of events (for example only push events)

Graininess of those settings in GitHub is not so high. Basically, it's only 2 options: 1) enable/disable push events and 2) enable/disable all the rest. Here we want to cover the case when we want to process only push events and a subset of all other events (for instance: sync repos and have published by the bot comments, but don't run pipelines for PRs).

> also, and most importantly, it won't allow to enable/disable features that depend on the same type of webhook. For example running the pipelines or doing cherry-picks all work based on commits.

You are absolutely right here. The idea here is to make a first step in a direction of managing features separately (enabling/disabling/configuring) without significant refactoring. To me it sounds like the variable names deserves for a change to be explicit that it's events filtering and not features management. @lluiscampos what do you think?

0lmi:
> What CFEngine needs is to start pipelines on PR for that repos (which is not happening with the current implementation).

@lluiscampos as far as I understood from the chat with @olehermanse this is not needed by CFEngine team

olehermanse:
@lluiscampos @0lmi As a user, something like this would make sense to me:

```
start_deployment_pipeline_on_push_or_merge = true
sync_repo_with_gitlab = true
run_CI_pipeline_on_pull_requests = false
post_comments_in_github = false
read_and_respond_to_comments_in_github = false
```

Not saying you have to implement it like that, but that is roughly what makes sense to me as an end user.

0lmi:
With the current events filtering and `SYNC_REPOS_LIST` the customisation can be the following:

```
sync_repo_with_gitlab = true/false
run_CI_pipeline_on_pull_requests && post_comments_in_github = true/false
read_and_respond_to_comments_in_github = true/false
```

In our case `sync_repo_with_gitlab` is equals to `start_deployment_pipeline_on_push_or_merge`, because as far as `.gitlab-ci.yml` exists in a repo GitLab starts pipelines by itself and it's out of our control.

So, with the events filtering it's almost perfect according to Ole features segregation.

olehermanse:
> In our case sync_repo_with_gitlab is equals to start_deployment_pipeline_on_push_or_merge, because as far as .gitlab-ci.yml exists in a repo GitLab starts pipelines by itself and it's out of our control.

Yep, I know, but from my perspective this is more like an "implementation detail". When I create a repo, I just want to say the things (CI features) that I want in the repo, and I don't care that much how it is achieved by bot or GitLab or something else. We could even imagine a system where the bot would manage settings of repos in GitLab, but that's probably not very high up on the priority list.

The point of my comment above was mainly to communicate what it looks like from my perspective, not to tell you to do it exactly like that.

lluiscampos:
@olehermanse Let's discuss it f2f for me to better understand the scenario. I'll ping you later today in the office.

main.go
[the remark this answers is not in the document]
lluiscampos: I have a different case, what if a feature requires a both a push and a PR hook? For example I would consider that "GitLab integration" feature consist on both pipelines on PRs and sync on push (This is related to the pending discussion I have with @olehermanse ).

With such a case then splitting by type of hook is not the first step but rather a side step that we'll need to revert later.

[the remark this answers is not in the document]
0lmi: The use case you are describing is actually covered by events filtering.

The only use case, which is not covered is when you want to enable PR hook but completely disable all comments from bot. If PR hook event is enabled, then automatically you get warnings if the bot fails to call GitLab API and a hint about interaction with the bot.

0lmi:
To summarise (and maybe simplify): the current CFEngine team's need is only sync repos on pushes.

main.go
```diff
@@ -85,16 +90,30 @@ const (
...
-	logrus.SetLevel(logrus.InfoLevel)
+	//
+	// Currently we don't have a distinguishment between GitHub events and features.
+	// Different features might be implemented across different events, but in future
+	// it's probability that we might implement proper features selection. For now the
+	// straight goal is to being able to configure the runner to only sync repos on
+	// push events and disable all the rest (to be used by the CFEngine team).
+	//
+	// default: process push events (sync repos) if not explicitly disabled
+	isProcessPushEvents := os.Getenv("DISABLE_PUSH_EVENTS_PROCESSING") == ""
```
lluiscampos, line 109: I guess it is the same, but we could use something like
```suggestion
	_ , isProcessPushEvents := os.LookupEnv("DISABLE_PUSH_EVENTS_PROCESSING")
```

In both cases it has the gotcha that setting the variable to `false` or some other negative meaning string still validates to "true". Anyway as we will redesign this part is okey to keep it simple for now :+1:

main_push.go
```diff
@@ -19,11 +19,23 @@ func processGitHubPush(
...
 
-	err := syncRemoteRef(log, repoOrg, repoName, refName, conf)
-	if err != nil {
-		log.Errorf("Could not sync branch: %s", err.Error())
+		err := syncRemoteRef(log, repoOrg, repoName, refName, conf)
+		if err != nil {
+			log.Errorf("Could not sync branch: %s", err.Error())
+			return err
+		}
 	}
```
lluiscampos, line 30: I would add here also another debug log entry, like "not syncing repository ______".

Or, even more accurate, restore the `log.Debugf("Got push event :: repo %s :: ref %s", repoName, refName)` to where it was (so that it is printed every time there is a push event, as the message suggests), and here only another message like "repo ______ not being watched, ignoring"

README.md
```diff
@@ -2,21 +2,20 @@
...
-
-Set `WATCH_REPOS_PIPELINE` for the list of repositories for which to run
-`mender-qa` Pipeline. See source code for defaults.
-
-This is a Mender specific feature.
-
 ### GitHub -> GitLab sync
 
-All repositories from the configured GitHub organization are synced with GitLab.
+By default all repositories from the configured GitHub organization are synced with GitLab if `SYNC_REPOS_LIST` (comma separated values) is not set.
```
lluiscampos, line 7: I suggest to be here more explicit in what the variable actually does.

For example:
```suggestion
By default all repositories from the configured GitHub organization are synced with GitLab. To select a subset of repositories to sync, set `SYNC_REPOS_LIST` env variable with a comma separated list of repositories.
```

The issue I see with the current text is that it really doesn't say what the variable does, only that it won't sync all repos -> but what repos then will be synced?

lluiscampos (approved):
Great work. Thank you for your patient and being such a good mediator between @olehermanse and me :+1: 

Some suggestions for minor improvements, but it looks ready to go!

lluiscampos (approved):

main.go
0lmi: This is a little bit tricky, because it sets reverse value. So, if the variable is empty then we process the event. This is needed to keep default behaviour with no extra settings for the current deployments. With `os.LookupEnv` the value is not reversed and additional settings will require for the current deployments to keep the events enabled.

lluiscampos: Thank you for the explanation.

PR #212, opened 2023-01-31
fix: retry the pipeline start in case of errors

oleorhagen:
This should help mitigate some of the issues we are seeing starting pipes, when
the base branch is not (yet) present.

Ticket: QA-519
Changelog: None

### DIFF
--- a/main_pullrequest.go
+++ b/main_pullrequest.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/gin-gonic/gin"
 	"github.com/google/go-github/v28/github"
@@ -19,6 +20,28 @@ var (
 	changelogPrefix = "Merging these commits will result in the following changelog entries:\n\n"
 	warningHeader   = "\n\n## Warning\n\nGenerating changelogs also resulted in these warnings:\n\n"
 )
+
+type retryParams struct {
+	retryFunc func() error
+	compFunc  func(error) bool
+}
+
+const (
+	doRetry bool = true
+	noRetry      = false
+)
+
+func retryOnError(args retryParams) error {
+	var maxBackoff int = 8 * 8
+	err := args.retryFunc()
+	i := 1
+	for i <= maxBackoff && args.compFunc(err) {
+		err = args.retryFunc()
+		i = i * 2
+		time.Sleep(time.Duration(i) * time.Second)
+	}
+	return err
+}
 
 func processGitHubPullRequest(
 	ctx *gin.Context,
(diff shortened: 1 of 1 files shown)


main_pullrequest.go
```diff
@@ -70,18 +88,30 @@ func processGitHubPullRequest(
...
+					case !re.MatchString(err_.Error()):
+						log.Errorf("failed to start pipeline for PR: %s", err)
+						return false
+					default:
+						return true
+					}
+				},
+			})
 			if err != nil {
-				// post a comment only if GitLab is supposed to start a pipeline
```
0lmi, line 75: IMO having comments is a plus 🤷‍♂️

0lmi (approved):
LGTM 👍
IMO it would be nice to keep comments, but don't think it's a blocker.

main_pullrequest.go
```diff
@@ -70,18 +88,30 @@ func processGitHubPullRequest(
...
+						"No stages / jobs for this pipeline")
+					switch {
+					case err_ == nil:
+						return false
+					case re.MatchString(err_.Error()):
+						log.Infof("start pipeline for PR '%d' is skipped", pr.Number)
+						return true
+					case !re.MatchString(err_.Error()):
+						log.Errorf("failed to start pipeline for PR: %s", err)
+						return false
```
lluiscampos, line 106: Aren't these swapped? :eyes: 
```suggestion
					case re.MatchString(err_.Error()):
						log.Infof("start pipeline for PR '%d' is skipped", pr.Number)
						return false
					case !re.MatchString(err_.Error()):
						log.Errorf("failed to start pipeline for PR: %s", err)
						return true
```
Otherwise I don't get it.

lluiscampos, line 106: Also this second case could very well be merged with the `default`.

lluiscampos (changes requested):
Beautiful :star_struck: Although I think it is wrong...

main_pullrequest.go
oleorhagen: No, they're not. They return the value to the comparator, and not `retry`. But I agree. This is retarded, and difficult to read. I'll switch it up in the `retry` function itself, to return a type I think, with something like a named type of retry, and not.

Thanks for the feedback, I was a little unsure of this one.

oleorhagen: Good point, fixing!

main_pullrequest.go
oleorhagen: Yes, I agree. It just didn't really fit in anywhere after I moved things around.

main_pullrequest.go
oleorhagen: Pushed a fix which should make this a lot clearer now

oleorhagen: https://github.com/mendersoftware/integration-test-runner/commit/05fb4489596fa4dcb82462cdc1c5aab56de26b90

lluiscampos (approved):

PR #213, opened 2023-02-01
chore: Logging improvement

0lmi:
The intention of the change is to improve logging to reduce amount of log entries, which will contribute to infra costs savings and making work with logs more comfortable.

The idea is to use `/_health` endpoint for the healthchecks by a load balancer and don't log those calls.

### DIFF
--- a/main.go
+++ b/main.go
@@ -262,7 +262,14 @@ func doMain() {
 	logrus.Infoln("using settings: ", spew.Sdump(conf))
 
 	githubClient = clientgithub.NewGitHubClient(conf.githubToken, conf.dryRunMode)
-	r := gin.Default()
+
+	r := gin.New()
+	r.Use(
+		gin.LoggerWithWriter(gin.DefaultWriter, "/_health"),
+	)
+	if logrus.GetLevel() == logrus.DebugLevel || logrus.GetLevel() == logrus.TraceLevel {
+		r = gin.Default()
+	}
 	r.Use(gin.Recovery())
 
 	// webhook for GitHub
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -262,7 +262,14 @@ func doMain() {
 	logrus.Infoln("using settings: ", spew.Sdump(conf))
 
 	githubClient = clientgithub.NewGitHubClient(conf.githubToken, conf.dryRunMode)
-	r := gin.Default()
+
+	r := gin.New()
+	r.Use(
+		gin.LoggerWithWriter(gin.DefaultWriter, "/_health"),
```
oleorhagen, line 274: Hmm, I'm not completely following, so if I'm running at the `info` level, then what will this show? Will it still ignore `_health`?

main.go
```diff
@@ -207,16 +207,22 @@ func processGitHubWebhook(
 		if conf.isProcessPREvents {
 			pr := webhookEvent.(*github.PullRequestEvent)
 			return processGitHubPullRequest(ctx, pr, githubClient, conf)
+		} else {
```
oleorhagen, line 210: Nice!

oleorhagen (approved):

main.go
0lmi: Yes, the idea is to ignore logging for `_health` endpoint for log levels above `debug` and only have them for `debug` and `trace`.

oleorhagen: Understood, thx :+1:  I was a bit lazy, but I see now that `_health` is the explicit filter to the logger.
Small suggestion, create a string locally, with `filter := "_health"` and pass it into the `gin.LoggerWithWriter`?

0lmi: Good point! Thanks!

oleorhagen (approved):

PR #214, opened 2023-02-02
feat: Also comment on cherry-picking to the latest released version

oleorhagen:
We don't always know if the latest version will need support in the future or
not, so we really have to cherry-pick to it, just in case.

Ticket: QA-496
Changelog: None

### DIFF
--- a/cherrypicks.go
+++ b/cherrypicks.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"os/exec"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"text/template"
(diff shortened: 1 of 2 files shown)


oleorhagen:
Won't pass CI in some of the old tests due to the changes, which brings up the question of the integration version used in unit tests...

I know we've been over this before, but hardcoding it could be an option for our tests to get a known version.. But then again, has some weaknesses too. 

What are your thoughts on the topic these days @lluiscampos ?

oleorhagen:
I know the inline testdata is gigantic. You want it split out in a separate file?

oleorhagen:
@lluiscampos more to your liking?

lluiscampos (approved):
Looks good! :rocket:

PR #224, opened 2023-02-27
fix: Do not process edited pull requests

oleorhagen:
Dependabot sometimes does a lot of edits in a short timespan. This causes our
test-bot to try and sync the branch on every request, basically doing a simple
DDos of our branch syncing to GitLab.

The fix is simple. Just ignore edits to PR's, which really is the right thing to
do anyways.

Changelog: None
Ticket: QA-538

### DIFF
--- a/main_pullrequest.go
+++ b/main_pullrequest.go
@@ -72,7 +72,7 @@ func processGitHubPullRequest(
 
 	log.Debugf("Processing pull request action %s", action)
 	switch action {
-	case "opened", "edited", "reopened", "synchronize", "ready_for_review":
+	case "opened", "reopened", "synchronize", "ready_for_review":
 		msgDetails := "see <a href=\"https://console.cloud.google.com/kubernetes/" +
 			"deployment/us-east1/company-websites/default/test-runner-mender-io/logs?" +
 			"project=gp-kubernetes-269000\">logs</a> for details."


mzedel:
Nooooooo, please don't! I rely on this to trigger a sync so often when the regular sync fails... Otherwise I would have to switch to a branch and push empty commits or whatnot when I might not even be on a computer...

lluiscampos:
> Nooooooo, please don't! I rely on this to trigger a sync so often when the regular sync fails... Otherwise I would have to switch to a branch and push empty commits or whatnot when I might not even be on a computer...

I was wondering how many of use relied in this "feature" :rofl: 

This was not intended to work this way in the first place. By removing the "process by edit" feature we will simplify the flow (opening the door for split the bot into per feature threads or individual bots) and save from unnecessary processing most of the times.

What about an explicit command for re-sync instead? Like `@mender-test-bot sync`? We could make that sync feature listens two two events: either an update in the code (push or force-push) and an explicit user command.

oleorhagen:
> > Nooooooo, please don't! I rely on this to trigger a sync so often when the regular sync fails... Otherwise I would have to switch to a branch and push empty commits or whatnot when I might not even be on a computer...
> 
> I was wondering how many of use relied in this "feature" rofl
> 
> This was not intended to work this way in the first place. By removing the "process by edit" feature we will simplify the flow (opening the door for split the bot into per feature threads or individual bots) and save from unnecessary processing most of the times.
> 
> What about an explicit command for re-sync instead? Like `@mender-test-bot sync`? We could make that sync feature listens two two events: either an update in the code (push or force-push) and an explicit user command.

Makes sense to me!

lluiscampos (approved):

main_pullrequest.go
```diff
@@ -117,7 +117,7 @@ func processGitHubPullRequest(
 				postGitHubMessage(ctx, pr, log, msg)
 			}
 		}
-
+	case "edited":
```
kacf, line 120: You need a fallthrough here too, otherwise it won't run for the above actions. Or call it twice.

kacf: Sorry, this was wrong, the lone "edited" fix is the only thing we need. Pre-approving, you know what to do!

kacf (approved):

PR #230, opened 2023-03-20
fix: Add backoff to the GitLab sync

oleorhagen:
This adds a simple backoff implementation to the GitHub <-> GitLab sync, through
the `git fetch --depth` parameter.

It will in turn try deeper and deeper depths until eventually fetching the whole
branch before syncing it.

Ticket: QA-538
Changelog: None


mzedel (approved):

gitlab_sync.go
```diff
@@ -54,10 +54,38 @@ func syncRemoteRef(log *logrus.Entry, org, repo, ref string, conf *config) error
 		return err
 	}
 
+	depths := []int{5, 10, 50, 100, 0} // 0 = infinite depth
+	for depth := range depths {
+		err := fetchCmd(ref, repo, state, depth)
+		if err != nil {
+			break
```
mzedel, line 61: would this maybe be:
```suggestion
		if err == nil {
			break
```
?

PR #254, opened 2023-06-28
feat: added support for main branch

oldgiova:
Alvaldi will use `main` as default branch but I found issues with the GH > GL sync, for example pipeline skipped after merging PRs to `main`.
I hope this PR would fix the issues.

### DIFF
--- a/cherrypicks.go
+++ b/cherrypicks.go
@@ -89,9 +89,9 @@ func suggestCherryPicks(
 		return nil
 	}
 
-	// ignore PRs if they don't target the master branch
+	// ignore PRs if they don't target the master or main branch
 	baseRef := pr.GetPullRequest().GetBase().GetRef()
-	if baseRef != "master" {
+	if baseRef != "master" && baseRef != "main" {
 		log.Infof("Ignoring cherry-pick suggestions for base ref: %s", baseRef)
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


gitlab_sync.go
```diff
@@ -20,6 +20,7 @@ func shouldStartPipeline(branchName string) bool {
 		"staging",
 		"production",
 		"hosted",
+		"main",
```
lluiscampos, line 23: This will fix the issue that you are describing :+1: 

The rest of the changes are on features that I think we have disabled for Alvaldi. Nevertheless is no harm to add `main` for the cherry-pick suggestions and some tests.

cherrypicks.go
```diff
@@ -89,9 +89,9 @@ func suggestCherryPicks(
 		return nil
 	}
 
-	// ignore PRs if they don't target the master branch
+	// ignore PRs if they don't target the master or main branch
 	baseRef := pr.GetPullRequest().GetBase().GetRef()
-	if baseRef != "master" {
+	if baseRef != "master" || "main" {
```
lluiscampos, line 94:
```suggestion
	if baseRef != "master" && baseRef != "main" {
```

lluiscampos (changes requested):
Nice initiative fixing it yourself!

There is an small bug in `cherrypicks` :eyes:

lluiscampos (approved):

PR #269, opened 2023-10-04
Revert "Clean-up unnecessary env variable from Dockerfile(s)"

lluiscampos:
This reverts commit 6864cb7f23c478379fb94da29c13a4b459df1041.


oldgiova:
If the env variable is really not necessary, maybe we should remove it from the code instead? Meanwhile, we could apply [this workaround](https://github.com/mendersoftware/sre-tools/pull/328) to not disrupt the workloads.

oldgiova (approved):

PR #341, opened 2025-01-06
chore: update golang-version and migrate deprecated code

danielskinstad:
The newer golang version in the `golangci/golangci-lint` image breaks the pipeline; attempting to migrate to a newer golang version here as well - which required to update some deprecated code


lluiscampos (approved):

lluiscampos:
Thank you!

PR #343, opened 2025-01-06
chore: bump golang-image versions

danielskinstad:
We need a newer golang version in the unit tests, because go now allows patch-versions for the go version in the go.mod - which some newer dependencies require.

We also need this https://github.com/mendersoftware/mendertesting/pull/358


danielskinstad:
This is mainly needed for: https://github.com/mendersoftware/integration-test-runner/pull/340, the version needs to be 1.23.4, which is not allowed in older go versions

lluiscampos (approved):

PR #350, opened 2025-02-05
SEC-1473: feat: Support for cryptographically sign Git commits

lluiscampos:
We opted for not modifying the `git` commands from the source code and instead just externally configure Git to always sign commits.

With this approach we let the different deployments of this software to opt-in the feature (by mounting the required secret) but let the software run without it otherwise.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN CGO_ENABLED=0 go build
 
 FROM golang:1.23.5-alpine3.21
 EXPOSE 8080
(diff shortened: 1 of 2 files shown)


lluiscampos:
These three PRs go together:
* https://github.com/mendersoftware/integration-test-runner/pull/350
* https://github.com/NorthernTechHQ/mystiko/pull/299
* https://github.com/mendersoftware/sre-tools/pull/398

oldgiova (approved):

PR #357, opened 2025-02-20
Qa 863 integration tests on demand

danielskinstad:
I have no idea if this works :sweat_smile:
Will be good with some fresh eyes on this


gitlab_sync_test.go
```diff
@@ -11,27 +11,27 @@ func TestStartPipeline(t *testing.T) {
 		branchName     string
 		expectedResult bool
 	}{
```
lluiscampos, line 13: :see_no_evil: ,
There is a formatting typo in this first commit: "Renamed the variable `commandStartPipeline` to `commandStartClientPipeline"

menderqa_pipeline.go
```diff
@@ -522,10 +415,10 @@ func getBuildParameters(
 			Value: &readHead,
 		})
 
-	return getClientBuildParameters(buildParameters, build)
+	return _getClientBuildParameters(buildParameters, build)
```
lluiscampos, line 418: :sob: ,
This is pretty ugly. Maybe rename to `getClientAcceptanceTestsParameters` or similar. Once we get the rid of Mender 3.7 we could clean this further and squash the two functions.

main_comment.go
```diff
@@ -147,6 +151,46 @@ func processGitHubComment(
 	return nil
 }
 
+func protectBranch(conf *config, branchName string, pipelinePath string) error {
+	// https://docs.gitlab.com/ee/api/protected_branches.html#protect-repository-branches
+	opt := &gitlab.ProtectRepositoryBranchesOptions{
+		Name: &branchName,
```
lluiscampos, line 185: `allow_force_push` is `false` by default but I would add it here explicitly for clarity.

main_comment.go
```diff
@@ -147,6 +151,46 @@ func processGitHubComment(
...
+	return nil
+}
+
+func syncProtectedBranch(
+	log *logrus.Entry,
+	pr *github.PullRequestEvent,
+	conf *config,
+	pipelinePath string,
+) (string, error) {
+	prBranchName := "pr_" + strconv.Itoa(pr.GetNumber()) + "_protected"
```
lluiscampos, line 211: We will not allow force-pushing, so by reusing the same branch name it won't be able to sync a second time. Let me explain:
* PR 123 launched with commit A
* command "start pipeline" -> the bot creates pr_123_protected and runs pipeline
* PR rebased or commit amended. Commit A'
* command "start pipeline" -> the bot tries to update pr_123_protected but GitLab force-push so it will fail

One solution is to delete the branch in advance if it exists. Another solution is to add a timestamp suffix into the branch name.

lluiscampos (approved):
> I have no idea if this works

I think it will!

Okey, admittedly I am also a bit rusty on this codebase but from what I remember your changes make sense and the code is correct (except for one thing, see below).

**We need an acceptance test**. If you already have the setup to run the tests locally, the best is that we issue the new command on a dummy PR, capture the webhook payload, and then run the test. (However I am pre-approving, read below)

Otherwise if you are feeling adventurous we can:
* Merge it Monday morning
* We test it and fix potential bugs. If shit hits the fan we revert.
* Only then capture the payload of the GitHub -> bot webhook and then add the test to lock the behavior.

This tool is used only internally and as long as we communicate to our peers about the update is okey if we break it for a little while.

main_comment.go
lluiscampos: I decided to approve the PR. We can follow-up on this on a separate PR once we have tested the basic functionality (so that we don't add complexity for the first try)

PR #358, opened 2025-02-24
Fix integration pipeline

danielskinstad:


pullreq_pipeline.go
```diff
@@ -112,7 +112,6 @@ func syncBranch(
 ) error {
 
 	repo := pr.GetRepo().GetName()
-	org := pr.GetOrganization().GetLogin()
```
danielskinstad, line 115: TBH, I have no clue why this doesn't work when doing the `start integration pipeline`

lluiscampos: I am not sure neither. [Here](https://github.com/google/go-github/blob/782d39f7fb968e87ab1665c0581a1cd7c2bfaed5/github/event_types.go#L1217) there is some indication that the organization might be empty, but it says that it should be filled if the webhook is from a repo within an organization (?).

Anyway using the conf is safe, as there we know exactly what the value will be.

lluiscampos (approved):
Nicely done with the test :rocket:

PR #359, opened 2025-02-25
Fix integration pipeline

danielskinstad:


main_comment.go
```diff
@@ -222,6 +223,9 @@ func syncProtectedBranch(
 		mainErrMsg := "There was an error syncing branches"
 		return "", fmt.Errorf("%v returned error: %s: %s", err, mainErrMsg, err.Error())
 	}
+	// Arbitrary sleep to ensure the branch is
+	// created before we protect it
+	time.Sleep(time.Duration(5) * time.Second)
```
danielskinstad, line 228: Wrt to this, I don't see a good way to `check` if a branch is ready. Checking if the branch exists wouldn't do much I think, because it was able to halfway-protect it already

PR #360, opened 2025-02-26
fix: don't return error on 404 when deleting protected branches

danielskinstad:
Changelog: None
Ticket: None


lluiscampos (approved):
As discussed privately, something else is going on because [I got 404 in here](https://github.com/mendersoftware/integration/pull/2727#issuecomment-2684328652) where the branch should actually exist from the previous run.

The fix in itself is good so approving :+1:

PR #373, opened 2025-04-09
QA-972: Add option to skip CI from PR title

alfrunes:
If a PR title starts with `[NoCI]` (case insensitive) the CI will not be started automatically.

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -66,6 +67,11 @@ var qemuBuildRepositories = []string{
 	"mender-gateway",
 	"mender-snapshot",
 }
+
+const (
+	KiB = 1024
+	MiB = 1024 * KiB
+)
 
 const (
 	gitOperationTimeout = 30
@@ -282,6 +288,11 @@ func doMain() {
 	r.POST("/", func(context *gin.Context) {
 		payload, err := github.ValidatePayload(context.Request, conf.githubSecret)
 		if err != nil {
+			var mbErr *http.MaxBytesError
+			if errors.As(err, &mbErr) {
+				context.Status(http.StatusRequestEntityTooLarge)
+				return
+			}
 			logrus.Warnln("payload failed to validate, ignoring.")
 			context.Status(http.StatusForbidden)
 			return
@@ -314,7 +325,7 @@ func doMain() {
 
 	srv := &http.Server{
 		Addr:    "0.0.0.0:8080",
-		Handler: r,
+		Handler: http.MaxBytesHandler(r, 10*MiB),
 	}
 
 	go func() {
--- a/main_pullrequest.go
+++ b/main_pullrequest.go
@@ -215,6 +215,9 @@ func processGitHubPullRequest(
    <summary>my commands and options</summary>
    <br />
 
+   You can prevent me from automatically starting CI pipelines:
+   - if your pull request title starts with "[NoCI] ..."
+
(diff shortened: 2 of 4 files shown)


main.go
```diff
@@ -282,6 +283,11 @@ func doMain() {
 	r.POST("/", func(context *gin.Context) {
 		payload, err := github.ValidatePayload(context.Request, conf.githubSecret)
 		if err != nil {
+			var mbErr *http.MaxBytesError
+			if errors.As(err, &mbErr) {
```
merlin-northern, line 287: please double check, this is a pointer to a pointer now passed to As -- this is what you want?

main.go
```diff
@@ -314,7 +320,7 @@ func doMain() {
 
 	srv := &http.Server{
 		Addr:    "0.0.0.0:8080",
-		Handler: r,
+		Handler: http.MaxBytesHandler(r, 10*1024*1024),
```
merlin-northern, line 323: could we have a const or a var with some config at one point in the future?

main_pullrequest.go
```diff
@@ -73,6 +75,20 @@ func processGitHubPullRequest(
...
 		return nil
 	}
+	title := strings.TrimSpace(req.GetTitle())
+	subMatch := regexpTitleOptions.FindStringSubmatch(title)
+	var skipCI bool
+	if len(subMatch) > 1 {
+		titleOptions := strings.Fields(subMatch[1])
+		for _, option := range titleOptions {
+			switch strings.ToLower(option) {
+			case "noci":
```
merlin-northern, line 85: `const SubjectTagsNoCi = "noci"` ?

main_pullrequest.go
```diff
@@ -47,6 +47,8 @@ func retryOnError(args retryParams) error {
 	return err
 }
 
+var regexpTitleOptions = regexp.MustCompile(`^\[([^\]]+)\].+$`)
```
merlin-northern, line 50: I am not going to block it by this, but a simple tokenizer , or: 1st word with separator ']', second word from that with separator '[', and then split wrt to ',' would be easy, and support `[noci,fast,pass-please]` schema. up to you.

main_pullrequest.go
```diff
@@ -94,25 +110,27 @@ func processGitHubPullRequest(
...
-					case re.MatchString(compareError.Error()):
-						log.Infof("start client pipeline for PR '%d' is skipped", pr.Number)
-						return noRetry
-					default:
-						log.Errorf("failed to start client pipeline for PR: %s", compareError)
-						return doRetry
-					}
-				},
-			})
+			if !skipCI {
```
merlin-northern, line 113: perhaps add above ```if prRef != "" && !skipCI {...```?

main_pullrequest.go
alfrunes: Thought a simple regex would be easier to read, but I can write a simple function for parsing the title instead if preferable.

main_pullrequest.go
alfrunes: `prRef != ""` is already covered by the parent if branch.

main_pullrequest.go
alfrunes: I refactored the bit gathering the options. I'm using space separated options (`strings.Fields`) instead of comma separated if you don't mind.

merlin-northern (approved):
thank you. I only miss two things: a test case, and documentation, I would add it to the message of the bot as the place where it is most likely I will read it ;-)

main_pullrequest.go
merlin-northern: I meant to add the option check to the if `if prRef != "" {` above :)

PR #376, opened 2025-05-02
QA-1007: Replace source of truth for watch repos from release_tool to code

lluiscampos:
As per today, we have two opt-in pipelines that we can run through the bot, the integration pipeline (iow from `integration` repository) or the client pipeline (iow from the "client bundle" repositories).

Modify the code to reflect this:
* Disregard release_tool, use a hard-coded (but short) list of repositories instead.
* Rework the main comment to either offer the integration or the client pipeline as the main comment
* For suggesting cherry-picks, now we can only use the release tool on client repos. For others (mender-gateway at the time of writing) print a generic message instead.
* Note the TODOs added to the list of repos for further steps in the release split

I tried hard to add test for cherry-pick suggestions on non-client repos but it has proved non feasible. On the unit tests we do run real Git commands but we only access open sources repos and on acceptance tests we run Git commands with dry run option.


main_pullrequest.go
```diff
@@ -199,9 +199,16 @@ func processGitHubPullRequest(
...
-			"mentioning me and the command \""
+		// Two possible pipelines: client or integration
+		var botCommentString string
+		if pr.GetRepo().GetName() == "integration" {
+			botCommentString = `, start a full integration test pipeline with:
+   - mentioning me and ` + "`" + `start integration pipeline` + "`" + ``
+		} else {
+			botCommentString = `, start a full client pipeline with:
+   - mentioning me and ` + "`" + commandStartClientPipeline + `"`
+		}
```
alfrunes, line 210: Side-note: Maybe we could refactor the comment rendering using `text/template`? It's quite difficult to see how the final string is gonna end up reading this function.

cherrypicks.go
```diff
@@ -74,6 +76,51 @@ func getLatestReleaseFromApi(url string) ([]string, error) {
...
+
+	// get list of release versions
+	versions, err := getLatestReleaseFromApi(versionsUrl)
+	if err != nil {
+		return releaseBranches, err
+	}
+
+	repo := pr.GetRepo().GetName()
+	for _, version := range versions {
+		releaseBranch, err := getServiceRevisionFromIntegration(repo, "origin/"+version, conf)
```
alfrunes, line 102: Just double-checking; we're still supposed to use `release_tool.py` here, right?

main.go
```diff
@@ -57,17 +57,38 @@ var gitHubRepoToGitLabProjectCustom = map[string]string{
...
+
+// Repositories with opt-in pipelines
+var pipelineRepositories = append(clientPipelineRepositories,
+	"integration",
+)
+
+// LTS repositories for which to suggest cherry-picks.
+var ltsRepositories = append(clientRepositories,
+	"mender-gateway",
+)
```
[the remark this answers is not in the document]
lluiscampos: Please no :sweat_smile: 

On the surface it may seem like a good idea, but I am very concerned in that by adding a config file here we will very quickly end up mirroring the complexity of [the `components-maps`](https://github.com/mendersoftware/integration/blob/b79c95830263f151a56bca0f34cfe71c9127812b/component-maps.yml#L14) that we are moving away from. Why? Because we would need to mark which repos are client release, which are client build, which are lts, independent, whatnot... :exploding_head: I honestly think that this list with superset reads pretty nice.

Additionally, we already have some configuration scattered across environment variables (for example [which repos to sync](https://github.com/mendersoftware/integration-test-runner/blob/eca2603d9e367b4aa2bcbe47185bf0361d4109f8/main.go#L138)) so if any we could move everything into config file or the other way around.

main_pullrequest.go
lluiscampos: I will! Thanks

cherrypicks.go
lluiscampos: Yes, for the Client LTS "bundle" we still don't have a replacement for the version mapping and there we use the release tool. After the Gateway and tools releases, it will be time to modernize the client release too :sparkles:

lluiscampos:
Note to self,
FYI to the rest,

Don't merge this before [MEN-8338](https://northerntech.atlassian.net/browse/MEN-8338) so that we still run the legacy pipelines for Mender 3.7 there.

main.go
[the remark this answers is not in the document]
alfrunes: Ok, I can see it's definitely a conscious decision.
The only drawback is that we have to do another "release" whenever any of these hard-coded values change instead of being able to reconfigure and reuse for other projects (since the project is public).
Anyway, please dismiss my comment.

danielskinstad (approved):
Looks good to me

alfrunes (approved):

PR #382, opened 2025-05-30
chore: add mender-flash to client lts components

danielskinstad:
Ticket: QA-993


vpodzime (approved):
Would be nice to extend the commit message with some information about the impact of this change.

vpodzime (approved):
Nice, thanks!

PR #391, opened 2025-07-30
chore: remove 3.7.x legacy build

danielskinstad:
Since we no longer support 3.7.x it should be safe to remove the legacy build checks to trigger PRs. If we need to run the pipeline we could always run it manually or through the release-tool.

Ticket: QA-1116

### DIFF
--- a/menderqa_pipeline.go
+++ b/menderqa_pipeline.go
@@ -121,26 +121,7 @@ func getClientBuilds(log *logrus.Entry, conf *config, pr *github.PullRequestEven
 	return builds
 }
 
-func isLegacyBuild(build *buildOptions, buildParameters []*gitlab.PipelineVariableOptions) bool {
-	// We define a legacy build as either:
-	// * A PR in integration repo with baseBranch 3.7.x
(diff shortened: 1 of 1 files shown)


elkoniu (commented):
I am wondering if we will not need this again at any point in the future. The code we are working on right now may became "legacy" some day :) So maybe just move this to some "common / utils" library and we will be able to use it for some other checks in future?

danielskinstad:
@elkoniu Yes, our code right now will become legacy, but normally we don't tag mender-qa. What's referred to as "legacyBuild" here is the legacy release process. The reason mender-qa branched out for  3.7.x was because we split the release process and some breaking changes were introduced to the mender-qa pipeline as that was repurposed for the client-only releases. Previously we released everything at once; i.e. the server and the client was part of one bundle (the last all-encompasing bundle being 3.7) and was released together. Since then the server moved on to the monorepo (what's now mender-server/mender-server-enterprise) and defined their own release process, and the client has moved on to the "Client bundle" process which currently is 3.8.x, where we release all the client components that belongs to the LTS together. Perhaps we'll need something similar in the future, but I believe our goal should be to use mender-qa master, but who knows.

PR #404, opened 2025-09-22
ci: Upgrade to unit tests v2 template

lluiscampos:
Among other things, it will upgrade the golang image version from where to run unit tests.

The existing code is moved from `script` to `before_script` so that we can make better use of the templated code (like running the tests and collecting the coverage).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -12,7 +12,7 @@ include:
   - project: 'Northern.tech/Mender/mendertesting'
     file: '.gitlab-ci-check-golang-lint.yml'
   - project: 'Northern.tech/Mender/mendertesting'
-    file: '.gitlab-ci-check-golang-unittests.yml'
+    file: '.gitlab-ci-check-golang-unittests-v2.yml'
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -32,9 +33,6 @@ test:unit:
     - git clone --depth 1 https://github.com/mendersoftware/integration
     - export INTEGRATION_DIRECTORY="$(realpath integration)"
 
-    - go list ./... | grep -v vendor | xargs -n1 -I {} go test -v -covermode=atomic -coverprofile=../../../{}/coverage.txt {} 2>&1 | tee /dev/stderr | go-junit-report > ${CI_PROJECT_DIR}/test-results.xml || exit $?
```
elkoniu, line 35: This is not only update to v2 but also some `go` stuff removal:) I would propose to put it as two separated commits or add note in the PR that removed things are now just included in `v2` (my guess).

lluiscampos: It must be the same commit because the changes in the CI job are a consequence of the update. But you are right in that it is little justified so I am amending a bit the commit message 👍

PR #443, opened 2026-03-25
Feat/start previews

mzedel:
By going through the bot, only members of our Github org should be able to start the envs and they no longer need to be "developers" in the Gitlab org or attached to Gitlab at all.
To fully get this to into a usable state I'll follow up with:

- [x] ~defining credentials available for the test runner as the environment credentials OR allowing signups on the environment~ using hardcoded credentials for now as the environment is available via regular CI & on staging too
- [ ] running the tests based on the `RUN_REVIEW_E2E` - https://github.com/mendersoftware/mender-server/pull/1921

and to make it easier to consume:

- [x] posting the credentials ~to Slack~ as a comment


main_comment.go
```diff
@@ -462,3 +496,33 @@ func parseBuildOptions(commentBody string) (*BuildOptions, error) {
 
 	return buildOptions, err
 }
+
+func parseReviewAppEnterprise(commentBody string) bool {
```
alfrunes: Not super obvious what the return argument is here from the signature.
```suggestion
func parseReviewAppEnterprise(commentBody string) (isEnterprise bool) {
```

review_pipeline.go
```diff
@@ -0,0 +1,317 @@
...
+	gitlab "gitlab.com/gitlab-org/api/client-go"
+
+	clientgithub "github.com/mendersoftware/integration-test-runner/client/github"
+	clientgitlab "github.com/mendersoftware/integration-test-runner/client/gitlab"
+)
+
+const (
+	reviewDeployJobName        = "review:deploy"
+	defaultReviewAdminUsername = "mender-demo@example.com"
+	defaultReviewAdminPassword = "mysecretpassword!123"
```
pasinskim, line 20: Should we consider pulling these from env/config (adminUsernameKey/adminPasswordKey) so the eventual switch to real secrets doesn't require touching the code?

review_pipeline.go
```diff
@@ -0,0 +1,317 @@
...
+		commentBody := fmt.Sprintf(
+			"Review app deployment is not supported for repository `%s`.",
+			repoName,
+		)
+		comment := github.IssueComment{Body: &commentBody}
+		if err := githubClient.CreateComment(context.Background(),
+			conf.githubOrganization, repoName, pr.GetNumber(), &comment); err != nil {
+			log.Infof("Failed to comment on the pr: %v, Error: %s", pr, err.Error())
+		}
+		return fmt.Errorf("no review app config for repo %q", repoName)
```
pasinskim: For an unsupported repo we will have two error messages. One here `Review app deployment is not supported...` and then `Failed to start review app...`. Maybe we can trigger the error here and output to the user in the other place?

review_pipeline.go
```diff
@@ -0,0 +1,317 @@
...
+		"Found latest pipeline %d for ref %s in project %s",
+		latestPipeline.ID,
+		ref,
+		projectPath,
+	)
+
+	// List jobs in the pipeline and find the target job
+	jobs, err := client.ListPipelineJobs(projectPath, latestPipeline.ID, &gitlab.ListJobsOptions{
+		ListOptions: gitlab.ListOptions{
+			PerPage: 100,
```
pasinskim, line 71: Will this fail if we won't find the job within first 100? Should we paginate until the job is found?

.gitlab-ci.yml
```diff
@@ -1,6 +1,7 @@
 variables:
   DOCKER_REPOSITORY: mendersoftware/integration-test-runner
   DOCKER_VERSION: "27.3"
+  GO_VERSION: "1.25"
```
pasinskim, line 16: We are moving this from local `test:unit` scope to global scope. Previously only the unit-test job pinned 1.25; now every job that reads  GO_VERSION inherits it. If those jobs previously relied on a different/default Go version, this is an unintended behavior change.

mzedel: This is intentional + in a separate commit (https://github.com/mendersoftware/integration-test-runner/pull/443/changes/68c57d36ecf0d1938f08eeb3c89ac8bf08b6479a) - as the prior state would allow different versions between test + build - which at least for me caused an unexpected CI issue.

review_pipeline.go
mzedel: that's jobs per pipeline - the server enterprise pipeline currently has 50 jobs and is afaik among the biggest pipelines we have - IMO we should be good with 100 there...

review_pipeline.go
mzedel: 👍 yep, good point!

PR #445, opened 2026-04-08
fix(integration-test-runner): retry on protected branch push failure

p-targowicz:
Replace the delete-branch-and-recreate approach with an unprotect-push-reprotect cycle that retries up to 3 times when the force push fails due to GitLab API eventual consistency. This removes the arbitrary 5-second sleeps and instead only adds delay when the race condition actually occurs.

Ticket: QA-1411

### DIFF
--- a/main_comment.go
+++ b/main_comment.go
@@ -208,6 +208,40 @@ func protectBranch(conf *config, branchName string, pipelinePath string) error {
 	return nil
 }
 
+func unprotectBranch(conf *config, branchName string, pipelinePath string) error {
+	client, err := clientgitlab.NewGitLabClient(
+		conf.gitlabToken,
+		conf.gitlabBaseURL,
+		conf.dryRunMode,
+	)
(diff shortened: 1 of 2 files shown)


danielskinstad (approved):
You have two sign-offs in the commit. 
The code looks good to me. I am still not sure _why_ it sometimes fail, but let's hope this fixes it 🤞 

I am note sure if you've encountered this error yourself, but before marking the ticket as done, I would create a PR in integration and test it a bunch to ensure it works.

p-targowicz:
> You have two sign-offs in the commit. The code looks good to me. I am still not sure _why_ it sometimes fail, but let's hope this fixes it 🤞
> 
> I am note sure if you've encountered this error yourself, but before marking the ticket as done, I would create a PR in integration and test it a bunch to ensure it works.

thank you @danielskinstad ! removed duplicated sign off from commit

PR #446, opened 2026-04-14
chore: run PR pipelines on pull/NNN/merge revisions

p-targowicz:
chore: Switch all PR pipeline refs from pull/NNN/head to pull/NNN/merge. This tests PRs as if merged into the target branch, catching integration issues caused by interim changes before merge.

If the merge ref is unavailable (e.g. merge conflicts), the pipeline fails. Developers can manually trigger on /head at GitLab if needed.

Ticket: QA-1505


vpodzime (approved):
Looks good to me. My only question is:

> If the merge ref is unavailable (e.g. merge conflicts), the pipeline fails.

Does this mean that the pipeline starts and fails or (as I'd expect) it even fails to start? Is that properly communicated to the developer/PR?

p-targowicz:
> Looks good to me. My only question is:
> 
> > If the merge ref is unavailable (e.g. merge conflicts), the pipeline fails.
> 
> Does this mean that the pipeline starts and fails or (as I'd expect) it even fails to start? Is that properly communicated to the developer/PR?

Yes indeed ! For the integration repo sync path, the pipeline fails to start and you get a gh comment for other repos the pipeline starts and fails in Gitlab which also surfaced via a gh comment

danielskinstad (commented):
This should also be updated, right?: https://github.com/mendersoftware/integration-test-runner/blob/59524cd0b31e992f1c4eba0027200ae1d3714a46/main_comment.go#L475

I guess https://github.com/mendersoftware/mender-qa/blob/248a7dc3d6a751a42992093ba3b01bf87a0459cf/gitlab-pipeline/stage/init.yml#L104 will need changing as well

p-targowicz:
> This should also be updated, right?:
> 
> https://github.com/mendersoftware/integration-test-runner/blob/59524cd0b31e992f1c4eba0027200ae1d3714a46/main_comment.go#L475
> 
> I guess https://github.com/mendersoftware/mender-qa/blob/248a7dc3d6a751a42992093ba3b01bf87a0459cf/gitlab-pipeline/stage/init.yml#L104 will need changing as well

yes you're right @danielskinstad , well spotted ! added it and squashed commits

p-targowicz:
also found that `mender-qa` was not switched so switched it too https://github.com/mendersoftware/mender-qa/pull/870 or its beyond this ticket wdyt guys? can abort this change but I think we should move all

danielskinstad (approved):
> also found that mender-qa was not switched so switched it too https://github.com/mendersoftware/mender-qa/pull/870 or its beyond this ticket wdyt guys? can abort this change but I think we should move all

I think it's in scope

PR #447, opened 2026-04-28
fix: reword merge commit and handle stale merge ref

p-targowicz:
After switching to pull/NNN/merge refs two issues appeared:

1. GitLab displayed the auto-generated "Merge <sha>...." commit message

2. GitHub computes pull/NNN/merge asynchronously so the webhook can fire before the ref is updated added retry (up to 2 times, 5s apart)

Ticket: QA-1505

### DIFF
--- a/pullreq_pipeline.go
+++ b/pullreq_pipeline.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/google/go-github/v28/github"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


danielskinstad (commented):
Could we just get rid of the merge commit? `git reset --hard HEAD^2`

p-targowicz:
> Could we just get rid of the merge commit? `git reset --hard HEAD^2`

If we do so we'd end up with the same thing we had - running on `pull/NNN/head`, the question is if we want to do that ? imho if so it kill whole purpose of https://github.com/mendersoftware/integration-test-runner/pull/446

danielskinstad:
> the question is if we want to do that

No, ignore my comment -- I misunderstood what the command actually did

vpodzime (approved):
Thanks!

danielskinstad (approved):

alfrunes:
> If we do so we'd end up with the same thing we had - running on `pull/NNN/head`, the question is if we want to do that ? imho if so it kill whole purpose of #446

I'd prefer reverting #446 TBH. Now the bot will fail whenever there are merge conflicts between head and base of a PR in addition to making CI less predictable.

vpodzime:
> I'd prefer reverting #446 TBH. Now the bot will fail whenever there are merge conflicts between head and base of a PR in addition to making CI less predictable.

That's not a discussion to have in this PR, IMHO.

PR #449, opened 2026-04-30
fix(pullreq_pipeline): remove hardcoded git identity from amend command

p-targowicz:
The user.email and user.name -c flags override the git identity for the amend commit - it leads to sync issue  https://github.com/mendersoftware/integration-test-runner/pull/448#issuecomment-4343837109

Ticket: QA-1505


vpodzime (approved):
Looks legit to me.

PR #450, opened 2026-04-30
fix(golden-files): update git amend command in golden files

p-targowicz:
Update golden test files to reflect the removal of hardcoded git identity flags (-c user.email / -c user.name) in favour of -c commit.gpgsign=false.

Ticket: QA-1505


vpodzime (approved):
ACK

PR #451, opened 2026-04-30
fix(pullreq_pipeline): start GitLab pipeline after manual sync command

p-targowicz:
The @mender-test-bot sync comment handler only synced the branch to GitLab but never called startPRPipeline leaving the pipeline in skipped state (ci.skip was added when switching to pull/NNN/merge)

sett the organization field on the synthetic PullRequestEvent so that getRemoteURLGitLab resolves the correct GitLab group instead of failing with an unrecognized empty-string organization

Ticket: QA-1505

### DIFF
--- a/pullreq_pipeline.go
+++ b/pullreq_pipeline.go
@@ -203,8 +203,7 @@ func syncBranch(
 		return fmt.Errorf("%v returned error: %s: %s", gitcmd.Args, out, err.Error())
 	}
 
-	// Push but not don't trigger CI (yet)
-	gitcmd = git.Command("push", "-f", "-o", "ci.skip", "--set-upstream", "gitlab", prBranchName)
+	gitcmd = git.Command("push", "-f", "--set-upstream", "gitlab", prBranchName)
 	gitcmd.Dir = tmpdir
 	out, err = gitcmd.CombinedOutput()
 	if err != nil {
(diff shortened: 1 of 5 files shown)


mzedel (approved):

pullreq_pipeline.go
```diff
@@ -203,8 +203,7 @@ func syncBranch(
 		return fmt.Errorf("%v returned error: %s: %s", gitcmd.Args, out, err.Error())
 	}
 
-	// Push but not don't trigger CI (yet)
-	gitcmd = git.Command("push", "-f", "-o", "ci.skip", "--set-upstream", "gitlab", prBranchName)
+	gitcmd = git.Command("push", "-f", "--set-upstream", "gitlab", prBranchName)
```
danielskinstad, line 206: You said
> ci.skip was added when switching to pull/NNN/merge

But AFAICT this is pre-existing behavior and not something you added (see https://github.com/mendersoftware/integration-test-runner/commit/214e4a926b89b8d48d50f9181fa77e6b24a78cc0#diff-97fef263105739cfbbc69a2d7b7c7aaa762b0f006591f6e469e11b49864ec2e7R136)? The sudden change in behavior (i.e. pipelines not starting automatically when opening a PR) can't be explained by this, right?

danielskinstad (changes requested):
See comment below

pullreq_pipeline.go
p-targowicz: sorry for confusion and to get things straight - for me pipeline start on opening PR and on each commit - the thing im trying to fix is this `sync` thing that stopped working but i think I wont make it on time

PR #454, opened 2026-05-05
chore: add mender-client-subcomponents to client build list

danielskinstad:
So we can properly start mender-qa pipelines in PRs like https://github.com/mendersoftware/mender-client-subcomponents/pull/45


vpodzime (approved):
ACK

PR #458, opened 2026-05-21
 fix(integration-test-runner): keep branch permanently protected to avoid sync race condition

p-targowicz:
Fix race condition in branch protection sync.

The previous unprotect→push→reprotect cycle left the branch unprotected at the moment of the push, so pipelines triggered by that push ran without protected-branch CI secrets

Fix by keeping the branch permanently protected with AllowForcePush: true, so integration-test-runner can always force-push without ever unprotecting, 409 on protect is treated as success — branch is already protected

refactored protectBranch to accept an injected clientgitlab.Client and extracted syncProtectedBranchWithClient to add tsets


danielskinstad (commented):
Hey, this approach looks reasonable. A nit on the commit message: it only happened some times, i.e. there seems to be a race condition.
Thinking a bit out loud for a possible alternative:

IIRC, the reason for not having allow-force-push was to not allow any external user to force-push to a branch containing secrets.
However, imagine the flow:

1. external contributor creates PR
2. `pr_NR` is synced to gitlab
3. one from the mender team runs: `mender-test-bot start integration pipeline`
4. `pr_NR_protected ` is protected and created with allow-force-push set to true
5. external contributor force-pushed to his PR
6. `pr_NR` is synced to gitlab

The force push here doesn't mean any access for the external contributor to the `pr_NR_protected` branch in gitlab -- it should be safe to keep allow-force-push true. This means that we shouldn't need the unprotect-sync-protect flow: we could simply create the protection rule with allow-force-push and then sync the branch, right? With allow-force-push set to true, running `start integration pipeline` means that integration-test-runner would just force-push to the already-protected pr_NR_branch. That would hopefully simplify the logic a bit and also remove any protected/unprotected race conditions. 

Let me know if it makes sense and what you think.

p-targowicz:
> Hey, this approach looks reasonable. A nit on the commit message: it only happened some times, i.e. there seems to be a race condition. Thinking a bit out loud for a possible alternative:
> 
> IIRC, the reason for not having allow-force-push was to not allow any external user to force-push to a branch containing secrets. However, imagine the flow:
> 
> 1. external contributor creates PR
> 2. `pr_NR` is synced to gitlab
> 3. one from the mender team runs: `mender-test-bot start integration pipeline`
> 4. `pr_NR_protected ` is protected and created with allow-force-push set to true
> 5. external contributor force-pushed to his PR
> 6. `pr_NR` is synced to gitlab
> 
> The force push here doesn't mean any access for the external contributor to the `pr_NR_protected` branch in gitlab -- it should be safe to keep allow-force-push true. This means that we shouldn't need the unprotect-sync-protect flow: we could simply create the protection rule with allow-force-push and then sync the branch, right? With allow-force-push set to true, running `start integration pipeline` means that integration-test-runner would just force-push to the already-protected pr_NR_branch. That would hopefully simplify the logic a bit and also remove any protected/unprotected race conditions.
> 
> Let me know if it makes sense and what you think.

Yeah you're kinda right, simplified flwo - allow-force-push: true permanently , removed the post-sync unprotect+re-protect steps entirely also fixed the commit message to mention race condition 

@danielskinstad

main_comment.go
```diff
@@ -237,65 +217,60 @@ func unprotectBranch(conf *config, branchName string, pipelinePath string) error
...
-
-		if attempt < maxSyncRetries {
-			log.Warnf(
-				"Force push to %s failed due to branch protection (attempt %d/%d), retrying in %s",
-				prBranchName, attempt, maxSyncRetries, syncRetryDelay,
-			)
-			time.Sleep(syncRetryDelay)
-		}
+	// Unprotect first (idempotent — 404 ignored) so we can re-apply protection.
+	if err := unprotectBranch(client, prBranchName, pipelinePath); err != nil {
```
danielskinstad: Why do we need to re-apply the protection? I would think that we protect the branch once and since we have set allow-force-push, it should just work. One issue with that would probably be that protectBranch will return an error if the branch is already protected, but for that we could most likely just check if the status code is 409 (already protected) and then just continue to `syncer`.

p-targowicz: ok I agree there's no reason to protect now - just allow force push - move on on 409 -> push

sorry got lost after changes and should revise it after we agreed on previous change

client/gitlab/mocks/Client.go
```diff
@@ -13,11 +13,11 @@ type Client struct {
 }
```
danielskinstad, line 14: I just realized that this file is autogenerated by mockery, see the first line. You can run a command like: ```cd client/gitlab && mockery --name=Client --output=mocks```

danielskinstad (commented):
Apart from the mockery comment below it looks good to me

client/gitlab/mocks/Client.go
p-targowicz: done

danielskinstad (approved):
LGTM, thanks!

PR #460, opened 2026-05-28
chore: update yocto variables

danielskinstad:
Makes mender-test-bot start the mender-qa pipeline with the correct revisions after we've moved away from the poky repo.

Ticket: QA-1632


p-targowicz (approved):
lgtm

